PR #193, opened 2014-07-13
Setup cgroups on RHEL family

amaltson:
When I was installing and testing the Docker recipe on local test-kitchen CentOS and RHEL boxes, I found that even though Docker gets installed correctly, it actually isn't usable and fails on the first Docker pull because the cgroup mounts are properly configured. The workaround I discovered was to modify the cgroups recipe to work on RHEL family boxes and then execute it before the default docker recipe. With that in place, the binary installation works great.

This PR modifies the binary installation recipe to ensure cgroups are setup correctly, and modifies the cgroups recipe to also execute against RHEL family OSes.

Thank you for taking a look!

### DIFF
--- a/recipes/cgroups.rb
+++ b/recipes/cgroups.rb
@@ -1,14 +1,14 @@
 # TODO: Platforms handled here should be fixed in control_groups cookbook
 # Possibly: https://github.com/hw-cookbooks/control_groups/
-case node['platform']
-when 'oracle' || 'rhel'
+if platform_family?('rhel')
   package 'libcgroup'
 
   service 'cgconfig' do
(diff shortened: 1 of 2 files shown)


amaltson:
@tduffield I seem to have missed this entirely. Do you want both test-kitchen and chef spec tests?

tduffield:
For this change I think Chefspec tests would be sufficient.

amaltson:
Sorry, this feel off my todo list, will add test shortly.

someara:
Out of scope for the Docker cookbook... please set your cgroups with another cookbook

Issue #194, opened 2014-07-15
Stopping and removing a container doesn't work in docker 1.1.1


austenito:
When I run the following code, my docker containers aren't removed:

```
docker_container('foo') do
  action :stop
end

docker_container('foo') do
  action :remove
end
```

However, when I execute the following command manually, everything works as intended:

```
execute('stop container') { command "docker stop -t 60 foo" }
execute('remove container') { command "docker rm -f foo" }
```

Are the stop and remove actions not working how I think they should be?

PR #197, opened 2014-07-28
Using platform resources instead of shelling out

jschneiderhan:
Hi there. Thank you for sharing your excellent cookbook! I have been finding it very useful.

This PR is more for discussion than an actual merge candidate. I have been hitting an issue similar to #194 while trying to use the :redeploy action, and I think the problem is due to the timing between when platform resources are processed (embedded into the resource collection) vs when the shelled out commands are run. If the platform resources defined in the container provider executes immediately everything works fine. What I am seeing currently is the `docker rm` shellout execute prior to the `service` resource's `:stop` action, leading to the following error:

```
==> default: [2014-07-28T18:59:18+00:00] DEBUG: Platform ubuntu version 14.04 found
==> default:
==> default: ================================================================================
==> default: Error executing action `run` on resource 'execute[rm  a4117fb8ab6474f8d6d360345c96d3e13fa995175a89e7c2af2ce086232e6817]'
==> default: ================================================================================
==> default:
==> default:
==> default: Mixlib::ShellOut::ShellCommandFailed
==> default: ------------------------------------
==> default: Expected process to exit with [0], but received '1'
==> default: ---- Begin output of rm  a4117fb8ab6474f8d6d360345c96d3e13fa995175a89e7c2af2ce086232e6817 ----
==> default: STDOUT:
==> default: STDERR: rm: cannot remove 'a4117fb8ab6474f8d6d360345c96d3e13fa995175a89e7c2af2ce086232e6817': No such file or directory
==> default: ---- End output of rm  a4117fb8ab6474f8d6d360345c96d3e13fa995175a89e7c2af2ce086232e6817 ----
==> default: Ran rm  a4117fb8ab6474f8d6d360345c96d3e13fa995175a89e7c2af2ce086232e6817 returned 1
==> default:
==> default:
```

If I replace the shellout calls with `execute` resources (as shown below) all the commands process just as the code reads and work perfectly (for my use case, it very well may have broken every other one).

### DIFF
--- a/providers/container.rb
+++ b/providers/container.rb
@@ -119,7 +119,7 @@ def commit
     commit_end_args = new_resource.repository
     commit_end_args += ":#{new_resource.tag}" if new_resource.tag
   end
-  docker_cmd!("commit #{commit_args} #{current_resource.id} #{commit_end_args}")
+  execute "docker commit #{commit_args} #{current_resource.id} #{commit_end_args}"
 end
 
 def container_matches?(ps)
@@ -172,7 +172,7 @@ def container_name
 end
 
 def cp
-  docker_cmd!("cp #{current_resource.id}:#{new_resource.source} #{new_resource.destination}")
+  execute "docker cp #{current_resource.id}:#{new_resource.source} #{new_resource.destination}"
 end
 
 # Helper method for `docker_containers` that looks at the position of the headers in the output of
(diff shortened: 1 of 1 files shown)


someara:
Thanks for the PR.

In master, the provider now talks to the API directly instead of shelling out.

-s

PR #210, opened 2014-09-23
Correct comparison of default container names

benlangfeld:
A container resource with a service without an explicit `container_name` attribute was not idempotent. Checking the running state of the container by name requires lookup by the same name assumed at launch time, which in this case is some default value.

See https://github.com/bflad/chef-docker/issues/184#issuecomment-56533201


someara:
Thanks for the PR.

-s

PR #213, opened 2014-10-14
Adds rudimentary support for Arch Linux

logankoester:
Arch Linux

I am working in a VirtualBox, so I haven't executed the test suite against Arch Linux. Possibly some combination of attribute options will require further adjustment to work on this platform.

After making these changes everything I did test (by hand) is working quite happily.


someara:
Hello!

There was a pretty heavy refactor in 0.40.0.... I suspect chef-client should detect systemd and try to use it on Arch.

I'm going to close this issue for now, since this isn't going to merge cleanly.... if the new provider detection stuff in 12.4 doesn't work, you may need to add a line to the provider in libraries and send a new PR.

Thanks!

-s

PR #214, opened 2014-10-22
(On Ubuntu 12.04) Start Docker only after cgconfig started

fancyremarker:
On Ubuntu 12.04 without this change, the `docker` service starts before `cgconfig`, and I get the following error whenever starting a Docker container:

```
[error] container.go:815 Error running container: cgroup mountpoint not found for cpu
```

The only modification from the default `cgconfig` upstart script is that 

```
start on runlevel [2345]
```

has been changed to

```
start on (runlevel [2345] and starting docker)
```

Overall, this change doesn't feel quite clean, so I'm happy to edit is as you suggest.

### DIFF
--- a/recipes/cgroups.rb
+++ b/recipes/cgroups.rb
@@ -12,6 +12,13 @@ when 'ubuntu'
   package 'cgroup-bin'
 
   if node['platform_version'] == '12.04'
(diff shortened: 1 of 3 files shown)


someara:
Hello!

With the refactor, cgroup management is now out of scope for the Docker cookbook. Closing

PR #226, opened 2014-11-04
fixed the installation to run on centos 7.0.1406

vminkov:
1.the yum package is now (in centos 7) "docker" instead of "docker-io" http://stackoverflow.com/questions/25423840/how-to-install-docker-on-centos-7
2.the version line causes an error of the kind "no version specified, and no candidate version available" under centos

tested with  a docker centos7 image


someara:
Hello.

The 0.40.x refactor now only uses the binary installation method. Closing

PR #236, opened 2014-11-13
Example of running a container with ARGs

behemphi:
`Usage: docker run [OPTIONS] IMAGE [COMMAND] [ARG...]`

It took me quite a long time to figure out that the `[ARG...]` part of the command should be captured in this way.  I hope this will save others some time!

### DIFF
--- a/README.md
+++ b/README.md
@@ -552,6 +552,29 @@ docker_container 'docker-registry' do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


bflad:
@behemphi did the suggestion @tduffield provided satisfy your needs? If we need to make the documentation more clear, please let us know!

PR #247, opened 2014-12-05
fix the location from which settings are source when using upstart on el6

dpetzel:
Currently the upstart configuration is hard code to source its settings from `/etc/default/$UPSTART_JOB`, however the helper module method `docker_settings_file` returns `/etc/sysconfig/docker` so this is where all the `daemon_options` are written to. This has the effect that none of the daemon options are honored.

### DIFF
--- a/recipes/upstart.rb
+++ b/recipes/upstart.rb
@@ -7,6 +7,9 @@ template docker_upstart_conf_file do
   mode '0600'
   owner 'root'
   group 'root'
+  variables(
+    'docker_settings_file' => docker_settings_file
+  )
 end
 
 template docker_settings_file do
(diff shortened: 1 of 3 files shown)


dpetzel:
@tduffield  Is there a branch/fork somewhere I can point at to pick this fix up?

tduffield:
If you want to rebase this PR on the current master I will get it merged into master.

bflad:
This would be nice to fix! Could you please rebase and we'll get this pulled in?

dpetzel:
I might have some time this weekend to get to it. You won't hurt my feelings if just take the diff and apply it.

someara:
This should be working in master now. Closing

PR #250, opened 2015-01-06
Use double-quotes for DOCKER_OPTS

rchekaluk:
Minor generalization that allows interpolation of any embedded variables.


bflad:
LGTM -- I am adding a special note in the CHANGELOG for this change on the off-chance anyone is using an expected `$` in their DOCKER_OPTS that now needs to be escaped.

PR #254, opened 2015-01-10
Fixes #253 and #256 systemd: setting proper service name and port mappings

### LINKED ISSUE #253: Cannot enable/start docker service when using systemd on debian
The cookbook breaks when trying to configuring systemd service unit when running `systemctl enable docker` is missing `.service`

gosuri:
This is a fix for #253 and #256


gosuri:
I've noticed the .service for systemd service was only in older versions of system, so revered the changes from ba6cc21 and updated the PR

tduffield:
These fixes were merged elsewhere.

PR #251, opened 2015-01-10
Consistency between docker_image name and docker_container image attributes?

jperville:
I stumbled upon that one while trying to pull an docker image with a non-default (latest) tag and then use that same image to deploy a container.

Here is how the docker_image resource should be declared:

```ruby
docker_image 'busybox' do
  tag 'ubuntu-14.04'
  action :pull_if_missing
end
```

Here is how the docker_container container is declared:

```ruby
docker_container 'my-container' do
  image 'busybox:ubuntu-14.04'
  command 'sleep 1111'
end
```

At first, I declared my image resource same as the image attribute of the docker_container resource (it is DRY and follows the principle of the last surprise):

```ruby
docker_image 'busybox:ubuntu-14.04' do
  action :pull_if_missing
end
```

It appeared to work, except that the docker image was always being pulled (even if not missing) and the resource was always modified_by_last_action. This wasted around 3 seconds per docker_image resource and made my chef-runs longer than what it should be.

I found out later than the docker_image `tag` attribute has to be specified separately and not as part of the name attribute (if the tag is part of the name attribute then resource will instanciate however `#load_current_resource` won't detect the existing container).

@bflad @tduffield  would you be interested in a PR that splits the docker_image name attribute into image_name and image_tag attributes automatically? This would be backward compatible and would make it easier for developers to pull an image that has a non-default (latest) tag then deploy a container based on this same image.

### DIFF
--- a/providers/image.rb
+++ b/providers/image.rb
@@ -188,9 +188,9 @@ end
 
 def import
   if ::File.file?(new_resource.source)
-    execute_cmd("cat #{new_resource.source} | docker import - #{repository_and_tag_args}")
+    execute_cmd("cat #{new_resource.source} | docker import - #{repository_image_and_tag_args}")
   elsif ::File.directory?(new_resource.source)
-    execute_cmd("tar -c #{new_resource.source} | docker import - #{repository_and_tag_args}")
+    execute_cmd("tar -c #{new_resource.source} | docker import - #{repository_image_and_tag_args}")
   else
     import_args = ''
     if new_resource.image_url
@@ -201,7 +201,7 @@ def import
       import_args += new_resource.source
       import_args += " #{new_resource.image_name}"
     end
-    docker_cmd!("import #{import_args} #{repository_and_tag_args}")
+    docker_cmd!("import #{import_args} #{repository_image_and_tag_args}")
   end
 end
 
@@ -251,11 +251,10 @@ def remove
   docker_cmd!("rmi #{remove_args} #{image_name}")
 end
 
-def repository_and_tag_args
+def repository_image_and_tag_args
   docker_cmd_args = ''
   docker_cmd_args += new_resource.repository + '/' if new_resource.repository
-  docker_cmd_args += new_resource.image_name.split(':', 2).first # strip tag if any
-  docker_cmd_args += ":#{new_resource.tag}" if new_resource.tag
+  docker_cmd_args += image_and_tag_arg
   docker_cmd_args
 end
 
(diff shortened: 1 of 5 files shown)


jperville:
Here is my take at implementing the feature.

First, I had to fix the docker-image :tag action, which was broken because the `load_current_resource` method would only attempt tagging if a docker image with the specified tag already existed (because the `tag` attribute is ambiguous for that action).

Before, the following would never result in the 'busybox:mytag' tag being created, even if a 'busybox:latest' image was present.

```ruby
docker_image 'busybox' do
  tag 'mytag'
  force true
  action :tag
end
```

Then I added the ability to specify the tag of the existing image in the name attribute and it became possible to do the following:

```ruby
docker_image 'busybox:ubuntu-14.04' do
  tag 'mytag'
  force true
  action :tag
end
```

I have updated the README.md to the best I could to document the new (but still backward-compatible) behavior.

The docker-image action :tag remains tricky, but it should not be broken anymore. I didn't check if there are actions other than `:tag` which use the tag attribute both as tag of the existing image and tag of the new image, there might be some similar bugs left.

jperville:
The Rubocop failures are unrelated to this PR.

jperville:
The Rubocop failures are unrelated to this PR.

tduffield:
+1

@fnichol look good to you?

providers/image.rb
```diff
@@ -236,10 +253,9 @@ def remove
 
 def repository_and_tag_args
   docker_cmd_args = ''
-  if new_resource.repository
-    docker_cmd_args = new_resource.repository
-    docker_cmd_args += ":#{new_resource.tag}" if new_resource.tag
-  end
+  docker_cmd_args += new_resource.repository + '/' if new_resource.repository
+  docker_cmd_args += new_resource.image_name.split(':', 2).first # strip tag if any
```
bflad: Should this functionality exist in `image_and_tag_arg` instead? Do we need to remove `repository_and_tag_args` in favor of `repository_image_and_tag_arg`?

bflad:
I'm +1 for this, just a minor question about the image provider methods.

providers/image.rb
jperville: After checking, the `repository_and_tag_args` method could be simplified to leverage `image_and_tag_arg` and renamed to `repository_image_and_tag_arg` because the new name will be closer to what the method does. Want me to add to the PR?

bflad: If you have time, that would be fantastic! Thanks!

bflad:
@jperville if you have a few spare moments, would love to get this merged with the mentioned change. Otherwise please let me know and I'll finish it up. Thanks for all the contributions!

jperville:
@bflad almost forgot about this one, will have a look during the coming week-end.

jperville:
@bflad here is the last changeset, feel free to squash-merge if you like it. It may need rebasing.

someara:
Thanks for the PR.

-s

PR #257, opened 2015-01-12
Fixes repository resource execution

tupy:


tduffield:
The default action for `apt_repository` is `:add`. It shouldn't need to be specified explicitly. Can you maybe detail the issues that you are seeing?

PR #258, opened 2015-01-13
Fix checking docker container status on debian

fxposter:
`status_of_proc "$DAEMON" "$NAME"` will actually check that ANY docker process is running, not the needed one.


bflad:
Manually verified functionality on 1.5:

```
docker inspect -f '{{.State.Running}}' foobar
Error: No such image or container: foobar
docker inspect -f '{{.State.Running}}' jenkins
true
```

Thanks for the fix!

PR #259, opened 2015-01-13
Use registry on image inspection

fxposter:
When you are using custom registries `pull` command will not trigger notifications as expected because it will compare wrong ids.


bflad:
Thanks for this fix! LGTM.

PR #260, opened 2015-01-15
Fix accidental port changing when using systemd templates.

fxposter:
When you use `port.gsub!` - you are changing the value  port variable
itself, and as long as the value is not local to this method, but
is a part of new_resource record (when it's passed as an array or
string) - the previous code will modify `new_resource.port` instance
variable.


labria:
+1

PR #262, opened 2015-01-28
Added docker_client_user LWRP attribute (Fixes #261)

### LINKED ISSUE #261: Allow docker_registry to login for another account than root
When using `docker_registry` the `docker login` command is run as `root`. 
This is the expected behaviour when interacting with `docker` only through `chef`, but it doesn't allow another user of the provisioned node to interact with a registry.

The workaround is simple - copy the `/root/.dockercfg` to `$HOME/.dockercfg` - but it would be cleaner to allow a `user` parameter to the `docker_registry` resource.

omribahumi:


bflad:
I like the idea of this PR, but I think the attribute name may be misconstrued that this will run the `docker` command as the specified user, which this PR doesn't accomplish, but may actually be better behavior (assuming your client user is in the docker group).

Does it work if we instead add `:user => docker_client_user` instead to `shell_out`? I'm wondering if this would skip the need for messing with the HOME variable. Sorry I don't have time at the moment to test this myself.

omribahumi:
@bflad Regular `sudo` doesn't change the `HOME` environment variable (not sure how `:user` is implemented).
The way Docker detects the location of the docker config file is with the `HOME` environment variable.

The reason I changed this in the first place is the `HOME` environment variable is not set when running in `cloud-init`.

I think this is the right way of solving this. As for the name, it might be worth changing it to `config_username` or something.

bflad:
@tduffield @fnichol @someara do you have any opinions here? Much appreciated!

someara:
I've been standardizing on "run_user" and "run_group" across my work...
httpd and mysql as examples.

short and descriptive.

-s

Soliah:
Does this address the original issue of creating the docker config for each user? Or is the idea i'd run `docker_registry` specifying a different user for each?

someara:
Thanks for the PR.

-s

PR #263, opened 2015-01-28
Add additional_host attribute to container resource

fxposter:
It corresponds to `--add-host` CLI argument, described here:
https://docs.docker.com/reference/run/.


tduffield:
This looks good. Thanks!

PR #265, opened 2015-01-29
Access keyserver.ubuntu.com on port 80

Sauraus:
This resolves https://github.com/bflad/chef-docker/issues/264

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ tmp/
 *.sublime-project
 *.sublime-workspace
 *.un~
+.idea/
 
 # OS files
 .DS_Store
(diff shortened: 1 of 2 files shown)


Sauraus:
Those Travis errors already exist on trunk, which forked a few minutes before submitting my pull request.

tduffield:
So I'm not super convinced that port 80 should be the default. I think it would be better to make a note in the README about this and to offer an example about how you would override this in your local environment.

rtrauntvein:
I agree there @tduffield

gionn:
@tduffield @rtrauntvein state reasons instead of opinions, this is a good workaround for users behind restrictive firewalls and should cause no-harm to everyone else.

tduffield:
@Sauraus if you rebase this PR on master and Travis passes I'll merge it in. Given that Docker itself uses port 80 in it's installation procedures I no longer see a reason to enforce the default.

Sauraus:
@tduffield rebased and recommitted.

PR #268, opened 2015-02-11
Updated the /etc/init/docker.conf template.

jperville:
This PR brings our upstart docker service template up to date with the original at https://github.com/docker/docker/blob/master/contrib/init/upstart/docker.conf .

I found out about the issue when upgrading to lxc-docker 1.5.0; dpkg asked me if I want to replace the `/etc/init/docker.conf` file or keep my change. Ironically, the changes in the file are related to fixing an issue that I reported a while ago on the docker github that got merged last December.

### DIFF
--- a/templates/default/docker.conf.erb
+++ b/templates/default/docker.conf.erb
@@ -8,34 +8,51 @@ limit nproc 524288 1048576
 respawn
 
 pre-start script
-  # see also https://github.com/tianon/cgroupfs-mount/blob/master/cgroupfs-mount
-  if grep -v '^#' /etc/fstab | grep -q cgroup \
-    || [ ! -e /proc/cgroups ] \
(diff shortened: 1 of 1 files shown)


jperville:
In the long term, since the `lxc-docker` packages from docker.io have been including an upstart service configuration for a while, should we stop overriding that file in this cookbook with our own (potentially out of sync) version? I believe so but this should be the point of another PR.

jperville:
The travis fail is unrelated to this PR.

bflad:
Thanks for fixing this. The changes inside the upstart configuration LGTM (no missing dependencies or anything).

I completely agree with removing the template except for edge case binary/tarball installs, which I believe was the only reason this was implemented in the first place. There's probably a better way to approach the handling for only if we want to actually override the init_template and/or only binary/tarball installs, which I'd gladly accept a PR.

PR #269, opened 2015-02-12
Fix tests on Travis by following Rubocop style guidelines

fxposter:


bflad:
Thank you for this fix!

PR #270, opened 2015-03-02
Fixes BIP configuration errors by drastically shifting init script configuration

brianbianco:
Currently if you try to set the bridge ip via the docker defaults file you will get a hard failure.  This is caused by the  package installation of docker which automatically starts the service.  If you then stop the service and try to set the BIP (in the docker defaults file) it will explode saying that the BIP's do not match.

What I have done here is shift the init script recipe to appear BEFORE package installation.  I have also set the init scripts to take action :nothing by default.  Once the scripts are laid down and the package is installed, at the end of the default recipe it will look for the service resources and then set their actions to :start and :enable.

Other possible solutions:
  1. Setup init scripts to delete bridges before starting docker
  2. Have the package installation stop the service and delete the bridge it created upon installation
  3. When you install the deb use the argument `-o Dpkg::Options::="--force-confold"`

Either way this is sort of a problem, I know that v1 of the cookbook is being worked on.  Hopefully it doesn't suffer from this issue.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -15,14 +15,14 @@ provisioner:
 
 
 platforms:
-- name: centos-6.5
-  driver_config:
-    box: opscode-centos-6.5
-    box_url: http://opscode-vm-bento.s3.amazonaws.com/vagrant/virtualbox/opscode_centos-6.5_chef-provisionerless.box
-- name: centos-7.0
-  driver_config:
-    box: opscode-centos-7.0
-    box_url: http://opscode-vm-bento.s3.amazonaws.com/vagrant/virtualbox/opscode_centos-7.0_chef-provisionerless.box
+#- name: centos-6.5
+#  driver_config:
+#    box: opscode-centos-6.5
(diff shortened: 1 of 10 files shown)


brianbianco:
On a side note, I rewrote a bunch of the unit tests to reflect the new manner in which the services are started.   I was unable to get test kitchen to pass its tests, but I was also unable to get it to pass on the master branch so maybe that's just broken?

rtrauntvein:
+1 I have seen this issue come up, would be great to have a fix

bflad:
I haven't looked yet, but is this more suited for a conversation upstream? Honestly, I've always requested package maintainers to not start services automatically. IMHO I think its a bad practice, especially with services that may have side effects of running with a default configuration (such as this) and configuration management becoming more prevalent.

As for the unit tests, nice! Could you submit a separate PR with those? I'd rather those not get held up with this discussion.

bflad:
@brianbianco / @rtrauntvein any luck talking to the Docker team? If this PR is split up, I'd love to merge some pieces. Thanks!

someara:
This should be handled with the docker_service resource now.

PR #276, opened 2015-03-11
Added support for docker options device and cap-add

hvolkmer:
With this change it is possible to configure a container like this:
```
docker_container 'somecontainer' do
  device "/dev/fuse"
  cap_add "SYS_ADMIN"
end
```

which will allow in this example for mounting of images inside the container.


hvolkmer:
The build failed b/c of Rubocop which is fixed in https://github.com/bflad/chef-docker/pull/269/files

bflad:
LGTM!

PR #279, opened 2015-03-18
Allow docker_container memory to have String value (eg. memory='1G').

jperville:
This very modest PR allows String values for docker_container memory attribute.

Currently, the docker_container resource only allows `Fixnum` values for the attribute (in other words, memory value is implicitly in bytes) but the docker [documentation](https://docs.docker.com/reference/run/#runtime-constraints-on-cpu-and-memory) says that the `docker run --memory` parameter supports optional units. It is also possible to specify different memory and memory+swap limits.

With this PR I can set `docker.memory = '1G'` or `docker.memory = '150m,200m'`, while still allowing for Fixnum memory values as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,7 +517,7 @@ init_template | Template to use for init configuration | String | nil
 link | Add link to another container | String, Array | nil
 label | Options to pass to underlying labeling system | String | nil
 lxc_conf | Custom LXC options | String, Array | nil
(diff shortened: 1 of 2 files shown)


bflad:
Awesome! I'm sure a lot of people will appreciate this. Hopefully not too many folks using swap functionality. :wink:

PR #280, opened 2015-03-18
fix not_if statement

databus23:


bflad:
Interesting. Surprised this didn't bite sooner. Double checking the Chef documentation looks good to me: https://docs.chef.io/resource_common.html#not-if-examples

I'll pull in for next release.

PR #281, opened 2015-03-18
port function shouldn't return nil...

databus23:
... otherwise port.empty? raises a *undefined method `empty?' for nil:NilClass*


bflad:
Good catch. Thanks!

PR #285, opened 2015-03-18
fixes #284, runit finish script to stop a container

### LINKED ISSUE #284: Unstoppable container
After any container is deployed with chef-docker cookbook using:
```
docker_container 'my_container' do
  container_name 'my_container'
  image 'my_image'
  tty true
  detach true
  action :run
end
```
stopping it later manually (outside the cookbook) is impossible. I tried those commands:
```
docker stop my_container
sv stop my_container
```
Actually the latter command should be used, to make it work the following script is needed:  /etc/service/my_container/finish with contents:
```
#!/bin/bash

docker stop my_container
```
I think a finish script should be generated each time a docker container is created.

xmik:
See #284.

### DIFF
--- a/providers/container.rb
+++ b/providers/container.rb
@@ -415,6 +415,7 @@ def service_init
       supports :restart => true, :reload => true, :status => true, :stop => true
       action :nothing
       finish true
+      restart_on_update false
     end
   else
     service service_name do
--- a/providers/container.rb
+++ b/providers/container.rb
@@ -411,8 +411,10 @@ def service_init
   if new_resource.init_type == 'runit'
     runit_service service_name do
       run_template_name 'docker-container'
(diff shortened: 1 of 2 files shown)


xmik:
I think the runit service (chef resource) option: restart_on_update should be set to false, to not restart docker container when the runit script is updated. I will test it more. For now it failed with the data container, it tried to restart not running container.

xmik:
I tested on data container (not running as a daemon) and on not data container (running as a daemon) and it works for me. The virtual machine I tested on:
```
vagrant@vm:~$ cat /etc/debian_version 
7.5
vagrant@vm:~$ uname -r
3.14-0.bpo.2-amd64
```

bflad:
This seems good to me -- I'll get this in. Thanks!

PR #287, opened 2015-03-25
redhat 7 does not need the epel repository

databus23:
Redhat 7 ships the `docker` package in the `rhel-7-server-extras-rpms` repo.

See: https://docs.docker.com/installation/rhel/#red-hat-enterprise-linux-7-installation


bflad:
Sweet, thanks for this. I'll get this included with 0.37.0.

PR #288, opened 2015-03-26
Fix docker.socket unit for RHEL7

databus23:
The latest RHEL 7 docker rpm docker-1.4.1-37.el7 does not create a docker group anymore. This breaks the socker.socket unit file which defaults to *docker* for the socket group.

This post explains some details for not shipping a docker group by default for security reasons some more:
https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2015-January/msg00034.html


bflad:
Thanks for this submission! :+1:

PR #289, opened 2015-03-26
Update systemd service/socket files

databus23:
This updates the unit files to the latest state from upstream.

This pull request explains the details for the changes:
https://github.com/docker/docker/pull/8451


bflad:
Sweet! Makes sense. Thanks for posting the relevant discussion about the topic as well.

PR #290, opened 2015-03-26
ensure --host fd:// when using socket activation

databus23:
For socket activation to actually work the docker daemon needs to be started with `--host fd://` as all connections are handed via a file descriptor from the docker.socket unit.

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -82,7 +82,7 @@ EOH
         'dns' => Array(node['docker']['dns']),
         'dns-search' => Array(node['docker']['dns_search']),
         'exec-driver' => node['docker']['exec_driver'],
-        'host' => Array(node['docker']['host']),
+        'host' => node['docker']['init_type'] == 'systemd' ? ['fd://'] : Array(node['docker']['host']),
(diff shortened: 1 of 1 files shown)


bflad:
Great idea, although I think I'd prefer to handle setting `fd://` in the cookbook default attributes rather than here since theoretically you could not want `fd://` for some reason with systemd (I guess?). Reasoning is that this is what they do upstream by default:
https://github.com/docker/docker/blob/master/contrib/init/systemd/docker.service#L8

I'll PR `fd://` and the other upstream systemd changes in a few minutes. Thanks again for submitting this though!

databus23:
For historic purposes I repeat my statement from the commit comment here:

> I"m afraid this does not work they way you think it does. `node['docker']['host']` is also used to [populate the docker.socket unit file](https://github.com/bflad/chef-docker/blob/master/templates/default/docker.socket.erb#L6) with the "real" sockets that systemd should listen on. 
>
>I think semantically `node['docker']['host']` should always contain the actual sockets the user wants docker to be available on.  Hence my way of doing it in #290.

PR #292, opened 2015-04-10
readme formatting fix

wormzer:
A pipe needed escaping. Thanks for the cookbook -- great stuff!


bflad:
Great catch! Thanks so much!

PR #295, opened 2015-04-14
Rspec fixes mostly from #250 and #265 changes

bflad:
Running locally:
```
Pending: (Failures listed here are expected and do not affect your suite's status)

  1) docker::default when running on ubuntu 12.04 should make sure kernel >= 3.8
     # Not yet implemented
     # ./spec/default_spec.rb:32


Finished in 58.16 seconds (files took 1.93 seconds to load)
199 examples, 0 failures, 1 pending
```


bflad:
We're green again. :shipit:

PR #302, opened 2015-04-20
Fixed remaining `-notrunc` usages

docwhat:
The `-notrunc` flag causes a very annoying warning message to stderr.

This replaces the last usage (sysv init file), a comment, and the
minitests.


tduffield:
LGTM

PR #308, opened 2015-04-28
adding pid option support to container lwrp

mincedmit:
https://docs.docker.com/reference/run/#pid-settings-pid


someara:
Hi. 

The `docker_container` resource has been re-implemented to use the remote API directly... I went digging through the docker source trying to figure out how this actually works but conceded defeat for lack of glang experience.

I'm gonna close this PR for now. Feel free to send another one if you can figure it out.

I'd love to have an strace container =)

-s

Issue #309, opened 2015-05-04
Update container if attributes change


dpkp:
It would be useful to "redeploy" a container if any key attributes change:

for example, when working with containers that use release tags, it would be useful to do this:
```
docker_container 'foobar' do
  action :run
  image "foobar:#{tag}"
  detach true
end
```
and have chef converge the container to the correct image if the value of tag is changed (via attribute etc). the current behavior, as I understand it, is that changing the value of `tag` will not update a running container.

jagalan:
Agree that could be useful to have that. For now it can be solved by using the tag in docker_pull and then notifying docker_container with redeploy.

dpkp:
Hi jagalan -- unfortunately docker_pull + redeploy only works if the pulled image:tag changes . If, for example, the images are already available locally then chef does not mark the docker_image resource as changed, and so the :redeploy notification does not get sent to the container. So although redeploy works sort-of, it's not really a suitable work around.

wpietri:
A related issue I ran into is redeploying after I change a volume mapping. As far as I could tell, nothing happened when I added a new mapping, so I manually did `docker rm -f` and ran the chef client again. Not a big problem, but I was puzzled.

wpietri:
I'd like to solve this one, and I'll have some time to work on it the week of 8/3. If anybody has suggestions, I'm all ears.

dpkp:
my PR (#310) should fix. take a look and let me know if you have comments

yonah-codefresh:
I had a similar problem. Redeploy doesn't discriminate between containers that had/had not been changed. When the change was in a dependency like the image pulled, this could be solved with notifications but that meant using a nothing action by default. With a default of nothing, the containers wouldn't pick up changes to their own attributes.
What makes it harder to solve is there was no real code to get the parameters of a running container and I was too lazy to write it ;) Maybe now that the cookbook is speaking to the API it will be easier.
To solve it, I hashed new_resource.to_json on container start and saved to a file, then checked if the hashes matched before redeploy. If the hashes didn't match, I replaced the old hash and let the redeploy do it's work. It is still buggy so I'm sure there is a better way but maybe that helps.

someara:
Just a heads up.. I've started working on this. The state comparison is going to be tricky because there are various fields where the input and output end up in different formats. (ulimits, for example.)

jperville:
[...]

PR #310, opened 2015-05-04
stop/remove/run when container changed

dpkp:
See Issue #309 -- would love feedback on this approach.

### DIFF
--- a/providers/container.rb
+++ b/providers/container.rb
@@ -10,6 +10,8 @@ def load_current_resource
     @current_resource.created(ps['created'])
     @current_resource.id(ps['id'])
     @current_resource.status(ps['status'])
+    @current_resource.image(ps['image'])
+    @current_resource.command(ps['command'].gsub!(/^"|"$/, ''))
     break
   end
   @current_resource
@@ -94,7 +96,12 @@ action :restart do
 end
 
 action :run do
-  unless running?
+  if exists? && changed?
+    stop
+    remove_container
+    run
+    new_resource.updated_by_last_action(true)
+  elsif !running?
     if exists?
       start
     else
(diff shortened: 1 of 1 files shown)


CosmicCat:
I lifted this code. It was very needed. Had problems with this line:

return true unless (!new_resource.command.nil?) && (@current_resource.command.end_with? new_resource.command)

when i removed it, things seemed to work as I wanted them to.

I also had problems regarding images without specified versions. Worked around by explicitly specifying :latest in all cases. Which is probably good practice anyway.

dpkp:
what kind of problems did you have before removing that line? The intent there was to restart if the docker_container `command` changes.  I suspect that will be rare in most use cases, but I included for completeness.

CosmicCat:
new_resource.command was always empty. We are not supplying explicit commands to our containers and rather use the built in default. The default command is always displayed in docker ps.

to be clearer... i suspect that this only effects cases where the :command attribute on the docker_container resource is not being set.

CosmicCat:
which i guess means the .nil? isn't working as intended. hmm... now i am confused

dpkp:
there's no check if @current_resource.command is nil -- that could do it

dpkp:
failures here are rubocop style and dont apply to the code changes in this PR. would you like me to also cleanup the rubocop issues with the other code?

wpietri:
Ooh, I'm excited to see this moving ahead. One thing I don't understand: does this detect changes in environment and volume mapping? I'm not seeing that, but I'm glad to add it either here or in a separate PR.

someara:
Thanks for the PR.... mater has moved to talking directly to the API instead of shelling out, so this is invalid now. Closing for now.

PR #317, opened 2015-05-27
added --security-opt

fadeev2010:


someara:
Thanks for the PR.

-s

PR #319, opened 2015-05-28
Add support for the `--log-driver` option

grampajoe:
This adds the `--log-driver` option to the `docker_container` resource.

As an aside, would it make sense to add something like an `extra_options` attribute to pass arbitrary options to Docker? It might help in situations where there isn't explicit support for new options like `--log-driver`.


grampajoe:
It looks like the failing RuboCop check was added in a recent update, and isn't related to the changes here. Any thoughts on how to address that?

akerekes:
+1
I also would like to use this feature.

bkono:
I went to use the branch log-driver-support from @grampajoe and found there was a tiny problem with it (new_resources vs new_resource). Updated on my fork and pushed a pr back to fix it. Hopefully it can get merged back into this upstream after that fix.

rtrauntvein:
+1

habelson:
+1

akerekes:
+1

grampajoe:
I think this might still be held up because of unrelated failing Rubocop checks. Are there any plans to address those?

grampajoe:
It looks like this was added in 0.40.0. I'm closing this unless that's not the case.

PR #344, opened 2015-07-20
Add Tag to docker_container docs

jhmartin:
Add 'tag' to the docker_container reference docs. It defaults to 'latest'


someara:
Thanks for the PR.
This should be fixed up in master.

Issue #346, opened 2015-07-23
dns_search option for docker_service is not written correctly


fcoelho:
I've added a `docker_service`:

```ruby
docker_service 'default' do
  action [:create, :start]
  dns_search ['first.domain', 'second.domain']
end
```

The docs say that `dns_search` is supposed to be of type `String, Array`, but when I use an array (such as in the example above), the cookbook writes the docker command line as:

```
/usr/bin/docker -d --dns-search=["first.domain", "second.domain"] ...
```

when it should have been

```
/usr/bin/docker -d --dns-search=first.domain --dns-search=second.domain ... 
```

which leads to docker not starting since `["first.domain",` is not a valid domain name. If I try to use a simple string, chef fails with `Option dns_search must be a kind of Array!`

I'm running version 0.40.3 of this cookbook.

someara:
Hello!!

Thanks for reporting this... this should be fixed up in master and will be shipped in 1.0.0

-s

PR #347, opened 2015-07-23
The --dns-search argument appears multiple times

amaltson:
Before the code would output a Ruby Array in the Array syntax, so something like:

```
docker -d ... --dns-search=["custom.domain.com"] ...
```

The Array delimiters wasn't the [right format for the --dns-search argument](https://docs.docker.com/articles/networking/#dns) and the docker daemon would not start correctly. The daemon expects the --dns-search argument once per domain.

I've tested this on a local test-kitchen run in our wrapper cookbook and the daemon starts correctly :smile:


amaltson:
The build fails but the files aren't the ones I touched.

someara:
Hello!

Thanks for the PR... this has been fixed up in master and addresses #346 

closing!

-s

amaltson:
Awesome thank you!

PR #349, opened 2015-07-29
Changed version checking in attributes.

docwhat:
The `value_for_platform` takes version constraints (like bundle does).

This makes the version checks more robust than enumerating all possible
versions.

This should fix systems with RHEL/CentOS 6.7 (out this week).

[Documentation for value_for_platform](https://docs.chef.io/release/11-18/dsl_recipe_method_value_for_platform.html)

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -13,14 +13,14 @@ default['docker']['init_type'] = value_for_platform(
     'default' => 'sysv'
   },
   %w(redhat centos) => {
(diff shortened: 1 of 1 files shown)


docwhat:
I *think* this failed because the version of rubocop isn't the latest...

someara:
Thanks for the PR.... this cookbook is now resource based and no longer mandates the use of attributes... closing

PR #350, opened 2015-07-29
build-essentials is needed by test-kitchen

docwhat:


someara:
I'm not sure what was causing this to be the case... there has been significant rework in master and this isn't the case anymore closing

Issue #315, opened 2015-05-12
Error creating service template for systemd on Debian 8


alamut:
```
Chef::Exceptions::EnclosingDirectoryDoesNotExist
------------------------------------------------
template[/usr/lib/systemd/system/web-ui-dev.service] (/tmp/kitchen/cache/cookbooks/docker/providers/container.rb line 485) had an error: Chef::Exceptions::EnclosingDirectoryDoesNotExist: Parent directory /usr/lib/systemd/system does not exist.
```
That's the same story as with the Ubuntu 15.04 support.
There's no `/usr/lib/systemd/system` directory on debian 8 (jessie), the choices are `/lib/systemd/system` and `/etc/systemd/system`.

In my opinion, putting the service's definitions into `/etc/systemd` ideologically is more correct, since the `/lib/systemd` has the services provided by packages installed by the package manager.

Here is a workaround for those who's using 0.37.0 (since the paths are hardcoded), just add a correct only_if for your use case.
```
link '/usr/lib/systemd/system' do
  to '/etc/systemd/system'
  action :create
  only_if { platform? 'debian' }
end
```

The fix is trivial and is the same as for Ubuntu 15.04 within the `attributes/default.rb`, something like that (not tested).
```
default['docker']['systemd_system_dir'] = value_for_platform(
  'ubuntu' => {
    %w(15.04) => '/lib/systemd/system',
    'default' => '/usr/lib/systemd/system'
  },
  'debian' => {
    'default' => '/etc/systemd/system'
  },
  'default' => '/usr/lib/systemd/system'
)
```

lmickh:
With this fix, I still get an error when it tries to execute systemctl daemon-reload.

```
Errno::ENOENT: No such file or directory - /usr/bin/systemctl
```
Looks like the provider needs to have a condition for Debian to expect the file at /bin/systemctl instead.

someara:
This should be fixed up in master

PR #358, opened 2015-08-11
[GH-315] - Debian 8 systemd fix

lmickh:


someara:
Thanks for the PR.
This should be fixed up in master.
-s

PR #363, opened 2015-08-15
Update port attribute to allow an array to be passed in

### LINKED ISSUE #360: docker_container port attribute only acccepts a string argument
This should also accept an array of strings to handle multiple port mappings.

brandocorp:
This fixes #360


someara:
thanks for these!

PR #374, opened 2015-08-18
documenting selinux needs to be permissive or disabled

chasebolt:


jhmartin:
Does it need to be disabled, or is there some set of rules that can be applied to let chef-docker function with selinux?

chasebolt:
@jhmartin you can read up on the issue here https://github.com/docker/docker/issues/15498

PR #375, opened 2015-08-18
networkmode needs to be moved under hostconfig

### LINKED ISSUE #369: Network Mode Not Honored
Within a docker container the `network_mode` is not honored.
When setting `network_mode 'host'` and then using `docker inspect` you can tell `default` is set.
The solution is to move the `NetworkMode` setting in `provider_docker_container.rb` to be under `HostConfig`.

chasebolt:
resolves #369


someara:
Ha. We fixed this at the same time. Beat you to it =)

chasebolt:
hahaha until next time! ;)

Issue #380, opened 2015-08-20
DOCKER_OPTS string has a whitespace prepended to it


skmichaelson:
When executing the following code block:
```
docker_service 'docker' do
  host      ["tcp://0.0.0.0:4243", 'unix:///var/run/docker.sock']
  provider  Chef::Provider::DockerService::Upstart
  action    [:create, :start]
end
```

The result is as follows:
```
 * Documentation:  https://help.ubuntu.com/
New release '14.04.3 LTS' available.
Run 'do-release-upgrade' to upgrade to it.

Last login: Thu Aug 20 14:44:04 2015 from 10.0.2.2
vagrant@default-ubuntu-1204:~$ cat /etc/default/docker 
# Docker Upstart and SysVinit configuration file

# Customize location of Docker binary (especially for development testing).
DOCKER="/usr/bin/docker"

DOCKER_LOGFILE=/var/log/docker.log

# Use DOCKER_OPTS to modify the daemon startup options.
DOCKER_OPTS=" -H tcp://0.0.0.0:4243 -H unix:///var/run/docker.sock --ip-forward=true --log-level=info"
# FIXME


# If you need Docker to use an HTTP proxy, it can also be specified here.

# If you need Docker to use an HTTP proxy, it can also be specified here.


# This is also a handy place to tweak where Docker's temporary files go.
```

The line for DOCKER_OPTS has a whitespace prepended to it.

someara:
Merged and released as v1.0.7

PR #381, opened 2015-08-20
[GH-380] Remove prepended whitespce in DOCKER_OPTS

skmichaelson:
* Removes prepended whitespace in each string listed under docker_opts
* Uses #join to produce the docker_daemon_cmn
* Uses #join to render docker_opts in templates


someara:
Nice catch!

PR #388, opened 2015-08-21
removing retries as it is inherited from Resource class

### LINKED ISSUE #386: Incorrect password with docker_registry causes infinite retries & server lockout
When the password passed to the `docker_registry` resource is incorrect, there seems to be an infinite number of retries to log into the registry. After 10 tries, the server locks me out of attempting more logins for 1 hour, but the retries continue. Here's an example:

```
docker_service 'default' do
  action [:create, :start]
end

docker_registry 'https://index.docker.io/v1/' do
  username 'mrdziuban'
  password 'INVALID_PASSWORD'
  email 'email@email.com'
end
```

This is what `/var/log/upstart/docker.log` looks like:

```
Wrong login/password, please try again
time="2015-08-21T17:33:24Z" level=info msg="-job auth() = ERR (1)" 
time="2015-08-21T17:33:24Z" level=error msg="Handler for POST /auth returned error: Wrong login/password, please try again" 
time="2015-08-21T17:33:24Z" level=error msg="HTTP Error: statusCode=401 Wrong login/password, please try again" 
time="2015-08-21T17:33:24Z" level=info msg="POST /v1.16/auth" 
time="2015-08-21T17:33:24Z" level=info msg="+job auth()" 
time="2015-08-21T17:33:24Z" level=info msg="+job resolve_index(https://index.docker.io/v1/)" 
time="2015-08-21T17:33:24Z" level=info msg="-job resolve_index(https://index.docker.io/v1/) = OK (0)" 
time="2015-08-21T17:33:25Z" level=error msg="unable to login against registry endpoint https://index.docker.io/v1/: Wrong login/password, please try again"

...

time="2015-08-21T17:33:41Z" level=info msg="POST /v1.16/auth" 
time="2015-08-21T17:33:41Z" level=info msg="+job auth()" 
time="2015-08-21T17:33:41Z" level=info msg="+job resolve_index(https://index.docker.io/v1/)" 
time="2015-08-21T17:33:41Z" level=info msg="-job resolve_index(https://index.docker.io/v1/) = OK (0)" 
time="2015-08-21T17:33:41Z" level=error msg="unable to login against registry endpoint https://index.docker.io/v1/: Login: Too many failed login attempts. You have been locked out for one hour. Try again later or contact support. (Code: 429; Headers: map[X-Frame-Options:[SAMEORIGIN] Strict-Transport-Security:[max-age=31536000] Server:[nginx/1.6.2] Date:[Fri, 21 Aug 2015 17:33:41 GMT] Content-Type:[application/json] Vary:[Cookie]])" 
Login: Too many failed login attempts. You have been locked out for one hour. Try again later or contact support. (Code: 429; Headers: map[X-Frame-Options:[SAMEORIGIN] Strict-Transport-Security:[max-age=31536000] Server:[nginx/1.6.2] Date:[Fri, 21 Aug 2015 17:33:41 GMT] Content-Type:[application/json] Vary:[Cookie]])
time="2015-08-21T17:33:41Z" level=info msg="-job auth() = ERR (1)" 
time="2015-08-21T17:33:41Z" level=error msg="Handler for POST /auth returned error: Login: Too many failed login attempts. You have been locked out for one hour. Try again later or contact support. (Code: 429; Headers: map[X-Frame-Options:[SAMEORIGIN] Strict-Transport-Security:[max-age=31536000] Server:[nginx/1.6.2] Date:[Fri, 21 Aug 2015 17:33:41 GMT] Content-Type:[application/json] Vary:[Cookie]])" 
time="2015-08-21T17:33:41Z" level=error msg="HTTP Error: statusCode=500 Login: Too many failed login attempts. You have been locked out for one hour. Try again later or contact support. (Code: 429; Headers: map[X-Frame-Options:[SAMEORIGIN] Strict-Transport-Security:[max-age=31536000] Server:[nginx/1.6.2] Date:[Fri, 21 Aug 2015 17:33:41 GMT] Content-Type:[application/json] Vary:[Cookie]])"
```

This is happening at least on versions 1.0.7 and 1.0.8. I've confirmed on versions 0.37.0 and 0.43.0 that this is not an issue.

chasebolt:
setting retries in the hwrp fails. retries is inherited from the Resource class and defaults to a value of 0. resolves #386

### DIFF
--- a/libraries/provider_docker_container.rb
+++ b/libraries/provider_docker_container.rb
@@ -128,7 +128,6 @@ class Chef
 
       # Most important work is done here.
       def create_container
-        tries ||= new_resource.retries
         Docker::Container.create(
           'name' => new_resource.container_name,
           'Image' => "#{parsed_repo}:#{new_resource.tag}",
@@ -173,9 +172,6 @@ class Chef
             'VolumesFrom' => parsed_volumes_from
           }
         )
-      rescue Docker::Error => e
-        retry unless (tries -= 1).zero?
-        raise e.message
       end
 
       # Super handy visual reference!
@@ -198,23 +194,15 @@ class Chef
         next if c.info['State']['Restarting']
         next if c.info['State']['Running']
         converge_by "starting #{new_resource.container_name}" do
-          begin
-            tries ||= new_resource.retries
-
-            if new_resource.detach
-              new_resource.attach_stdin false
-              new_resource.attach_stdout false
-              new_resource.attach_stderr false
-              new_resource.stdin_once false
-              c.start
-            else
-              c.start
-              new_resource.timeout ? c.wait(new_resource.timeout) : c.wait
-            end
-
-          rescue Docker::Error => e
-            retry unless (tries -= 1).zero?
-            raise e.message
+          if new_resource.detach
+            new_resource.attach_stdin false
+            new_resource.attach_stdout false
+            new_resource.attach_stderr false
+            new_resource.stdin_once false
+            c.start
+          else
+            c.start
+            new_resource.timeout ? c.wait(new_resource.timeout) : c.wait
           end
         end
         new_resource.updated_by_last_action(true)
@@ -225,13 +213,7 @@ class Chef
         c = Docker::Container.get(new_resource.container_name)
         next unless c.info['State']['Running']
         converge_by "stopping #{new_resource.container_name}" do
-          begin
-            tries ||= new_resource.retries
-            c.stop
-          rescue Docker::Error => e
-            retry unless (tries -= 1).zero?
-            raise e.message
-          end
+          c.stop
         end
         new_resource.updated_by_last_action(true)
       end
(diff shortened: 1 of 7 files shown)


someara:
Aw geez. *facepalm*

So... I don't think removing it like this is the move... I ended up sticking it in there to get the test-kitchen runs to pass in the first converge. Image operations specifically, are super flakey on non-overlayfs backends because of $reasons that are addressed by the new dynamic binary builds.

Removing it completely means that its up to a recipe author to manually set retries above 0, which 
might not be the obvious thing to do.

Maybe we should rename it to api_retries?

-s

chasebolt:
ok so rename would work. but retries does still work, this isn't technically removing functionality. you can still set retries 3 on a container and it will retry it 3 times.

someara:
Right.

Renaming it means that it will retry 3 times by default, if the
docker_resource has the inherited "retries" set to (the default of) 0.

"retries 3" would end up retrying a total of 9 times.

-s

chasebolt:
my PR currently only fixes the retry functionality not being set. if we want to fix the original implementation of defaulting to 3 retries from the current infinite, then we will got the api_retries route.

before this PR
-retries defaults to 0 (infinite)
-setting retries to 3 does not work

after this PR
-retries defaults to 0 (no change)
-setting retries to 3 does work and ends up with 4 calls to the resource (original + 3 retries)

chasebolt:
oh ok i do see, now the retries defaults to 0 which means dont retry and you need that functionality for test kitchen. ill implement the rename instead unless i can find a way to set retries default

someara:
I renamed retries to api_retries in 1.0.11.

PR #391, opened 2015-08-23
Fix 'parsed_storage_driver' method error

Soulou:
```
[2015-08-23T22:20:17+00:00] ERROR: docker_service[default] (scalingo::docker line 12) had an error: NameError: No resource, method, or local variable named `storage_driver' for `Chef::Provider::DockerService::Upstart ""'
```


Soulou:
Great! :+1:

PR #392, opened 2015-08-23
Correctly define ip_forward flag, not present by default in opts

Soulou:
As it is defined to be a boolean in the resource, the only possible
values allowed to the user are true or false. Let docker use its default
value instead of setting a default flag.

If the user knows what he's doing he'll be able to give a specific value
to the flag.


Soulou:
Thanks

PR #393, opened 2015-08-23
No default flag for --log-level, let docker set the default value without specifying the flag

Soulou:


Soulou:
Thank you for merging

PR #406, opened 2015-08-28
adding ability to set docker api read write timeouts

### LINKED ISSUE #403: docker_image :pull - Docker::Error::TimeoutError read timeout reached
I get read timeouts occasionally when using docker_image to pull images. I don't see an option in the documentation or in the code for setting the read timeout.

I don't know if this is related, but I see this in the docker api documentation:

> These methods are prone to read timeouts.  Docker.options[:read_timeout]  will need to be made higher than 60 seconds if expecting a long time between events.

If this is causing the issue, could you make 60 seconds the default, but provide an attribute to override the default for pulls that take longer than 60 seconds.

My current workaround is to login to the host, pull the image, and change the docker_image to action to :pull_if_missing

chasebolt:
i split the api timeouts into two attributes matching the docker api. not sure if you want to combine this into one attrib or not, if you do let me know and i can change it.

resolves #403

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,14 @@ docker_image 'alpine' do
 end
 ```
 
+specify read/write timeouts
+```ruby
+docker_image 'alpine' do
+  read_timeout 60
+  write_timeout 60
+end
+```
+
 ```ruby
 docker_image 'vbatts/slackware' do
   action :remove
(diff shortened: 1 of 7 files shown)


chasebolt:
eh travis isnt very happy; the docker binary is failing to download.

kkeane:
It's not quite clear what unit to use - is the timeout in minutes or seconds? Also, could you add some documentation about what "read timeout" and "write timeout" actually mean? I tried a docker_image with action :build and set both read_timeout and write_timeout to high values (7200, assuming that it is seconds), but the build still timed out after little more than one minute. Before sending you on a wild goose chase, this could very well be an unrelated problem, but it's exacerbated by not understanding the details of the two timeout values.

chasebolt:
@kkeane these timeouts have been discussed in the docker ruby api. I have also tested these timeouts that they do have an affect when pulling a large image. i did not test building an image though, so that may need to be explored more if this didnt resolve your issue.

https://github.com/swipely/docker-api/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+timeout

chasebolt:
if building is still fails for you, please open a new issue with examples and output.

PR #437, opened 2015-09-18
Redeploy unstarted containers using action_create instead of action_run.

jperville:
This PR is my attempt at fixing #432.

For some reason, I was not able to run the serverspec ("kitchen setup" then "kitchen verify" are no-ops), so I ran my rspec examples manually on both resources-ubuntu-1404 and resources-ubuntu1504 instances after converging them.

My server spec example:
```ruby
# docker_container[unstarted_redeployer]

describe command("docker ps -af 'name=unstarted_redeployer$'") do
  its(:exit_status) { should eq 0 }
  its(:stdout) { should_not match(/Up/) }
  its(:stdout) { should match(/Created/) } if docker_version.to_f >= 1.8
end
```

The results (note the different container status pre and post docker 1.8).

On 14.04 (docker 1.6.2):
```sh
root@resources-ubuntu-1404:~# docker ps -a | grep redeployer
d8a8e41dc7e0        alpine:3.1               "nc -ll -p 7777 -e /   About a minute ago                                                                 unstarted_redeployer      
151eb96b1e3f        alpine:3.1               "nc -ll -p 7777 -e /   About a minute ago   Up About a minute             0.0.0.0:32777->7/tcp            redeployer                
```

On ubuntu 15.04 (docker 1.8.2):
```sh
root@vagrant:~# docker ps -a | grep redeployer
14fbef4889dc        alpine:3.1               "nc -ll -p 7777 -e /b"   5 seconds ago       Created                                                       unstarted_redeployer
dec1e2c6285d        alpine:3.1               "nc -ll -p 7777 -e /b"   5 seconds ago       Up 4 seconds                  0.0.0.0:32777->7/tcp            redeployer
```

PR #439, opened 2015-09-18
Qualify port bindings with protocol even when implicitly tcp

### LINKED ISSUE #427: docker_container and port syntax
I have noticed a difference related to port bindings between the containers created with the cookbook (currently version 1.0.24) and those created with an equivalent docker client command-line.

I create my container in the cookbook like this:
```ruby
docker_container 'my-app' do
  repo 'busybox'
  tag 'ubuntu-14.04'
  port ['80', '22']
  ...
end
```

I am trying to write an assertion in my cookbook test suite to check whether the container is publishing a given port. With version 0.37 of this cookbook (which used 'docker' client command-line instead of the API), I would just search for `80/tcp` in the output of `docker inspect --format='{{ .HostConfig.PortBindings }}' for example.

However, now that the containers are created using the docker API, I can only match for `80` (without 'tcp' at the end), which means that containers created by the cookbook and by the command-line API do not docker-inspect the same anymore.

Here is a sample shell session demonstrating the problem.

```sh
# when we start, we are supposed to have converged the node and the `my-app` container exists.
root@docker-ubuntu-1404:~# docker inspect --format='{{ .HostConfig.PortBindings }}' my-app
map[22:[{ }] 80:[{ }]] # <--- new format, no '/tcp' present
root@docker-ubuntu-1404:~# docker rm -f my-app
my-app
# now I recreate the container manually using equivalent command-line
root@docker-ubuntu-1404:~# docker create --name="my-app" --publish="22" --publish="80" busybox:ubuntu-14.04 /bin/top
WARNING: Your kernel does not support swap limit capabilities, memory limited without swap.
dd7e74178fc99a23e1781028b539c2f19ba039bab17c7856309389fdbd3fe856
root@docker-ubuntu-1404:~# docker inspect --format='{{ .HostConfig.PortBindings }}'
my-app
map[22/tcp:[{ }] 80/tcp:[{ }]]  # <-- note the presence of /tcp in the output.
```

Would it be possible to change the way port bindings are created in the cookbook to follow the convention of the docker command-line client?

jperville:
Fixes #427.

Validated by running `kitchen test resources-182-ubuntu-1404` (no time/bandwidth to run on all supported combinations).

### DIFF
--- a/libraries/_port_binding.rb
+++ b/libraries/_port_binding.rb
@@ -44,5 +44,7 @@ class PortBinding
     when 1
       @container_port = parts[0]
     end
+    # qualify the port-binding protocol even when it is implicitly tcp #427.
+    @container_port << '/tcp' unless @container_port.include?('/')
   end
 end
--- a/test/integration/resources-162/serverspec/assert_functioning_spec.rb
+++ b/test/integration/resources-162/serverspec/assert_functioning_spec.rb
@@ -141,16 +141,31 @@ describe command("[ ! -z `docker ps -qaf 'name=an_echo_server$'` ]") do
   its(:exit_status) { should eq 0 }
 end
 
+describe command("docker inspect --format '{{ range $port, $_ := .HostConfig.PortBindings }}{{ $port }}{{ end }}' an_echo_server") do
+  its(:exit_status) { should eq 0 }
+  its(:stdout) { should eq('7/tcp') }
+end
+
 # docker_container[another_echo_server]
 
 describe command("[ ! -z `docker ps -qaf 'name=another_echo_server$'` ]") do
   its(:exit_status) { should eq 0 }
 end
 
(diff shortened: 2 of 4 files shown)


someara:
I went to go test this, and I'm getting some errors

https://gist.github.com/someara/2066171ebcb5f3a95da5#file-gistfile1-txt-L590

-s

someara:
^ that's on Digital Ocean with an export KITCHEN_YAML=.kitchen.cloud.yml

jperville:
The last ones (about your sean_was_here) container are due to my changes in #437. Your test containers may be redeployed before it got a chance to reach 'running' state, it may work by adding some 'sleep' or maybe by changing the test assertion.

For the first one, it is a \n issue that I didn't spot on my only kitchen run, will try again and adjust the test if necessary (should_match will be more tolerant than eq).

someara:
just as long as kitchen test exists zero ;)

jperville:
Will submit separate PR in a bit.

jperville:
PR #442 submitted, the kitchen is all green again.

PR #442, opened 2015-09-18
Fixed serverspec after #439 and #437 merges.

jperville:
This PR fixes recent breakage in serverspec introduced by some of my recent PRs. Breakage log from @someara here: https://gist.github.com/someara/2066171ebcb5f3a95da5#file-gistfile1-txt-L590

First breakage was due using too strict matcher when parsing the docker inspect output in #439 
Second breakage was more serious and due to false positive when matching for 'unstarted containers' in #437


jperville:
Thank you. Automated tests are a good thing.

PR #443, opened 2015-09-21
Added docker_container log_driver and log_opts attributes.

jperville:
This PR allows to configure logging on a per-container basis, when the logging driver or its configuration is different than the systemwide (docker_service) defaults.

It introduces two new attributes for docker_container:
 - log_driver = %w( json-file syslog journald gelf fluentd none )
 - log_opts = Array of "key=value" strings that are specific to each log-driver
Which work similarly to the eponymous attributes of the `docker_service` resource.

Use the new attribute like this:
```ruby
docker_container 'syslogger' do
  command 'nc -ll -p 780 -e /bin/cat'
  repo 'alpine'
  tag '3.1'
  log_driver 'syslog'
  log_opts 'syslog-tag=container-syslogger'
end
```

The stubbed handling of docker_container log_config attribute has been fixed and updated to handle the 2 new attributes, which are an higher-level way to configure container logging.

I have tested compatibility with the `resources-162-ubuntu-1404`, `resources-171-ubuntu-1404` and `resources-182-ubuntu-1404` suites, so I think that it is compatible with all versions of docker that this cookbook supports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -810,6 +810,18 @@ docker_container 'api_timeouts' do
 end
 ```
 
+Specify a custom logging driver and its options
+
+```ruby
+docker_container 'syslogger' do
+  repo 'alpine'
+  tag '3.1'
+  command 'nc -ll -p 780 -e /bin/cat'
+  log_driver 'syslog'
+  log_opts 'syslog-tag=container-syslogger'
+end
+```
+
 #### Properties
 
 Most `docker_container` properties are the `snake_case` version of the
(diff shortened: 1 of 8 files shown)


jperville:
Please wait a little before merging, investigating #444.

jperville:
Thanks @someara . Still investigating, please don't do a supermarket release before I have fixed subtle redeploy issue.

someara:
Just saw your note about this... I was already in the process of releasing 1.0.30
I guess we'll release a 1.0.31 today too!

jperville:
About to write another issue for the new issue I found (similar to #444 but about exposed_ports, not about log_config). While I'm in there I want to fix it for good with another PR.

jperville:
Have another commit to append to this PR, if you can merge it.

someara:
I've already merged it... can you make a new PR?
-s

jperville:
Will do!

jperville:
More to come, have found a similar issue with volumes...

jperville:
@someara I have finished my group of PR, could you have a look an issue a new supermarket release if you like them all?

someara:
shall do!

PR #448, opened 2015-09-21
Fix unwanted redeploys due to calculation of exposed_port changes.

### LINKED ISSUE #446: docker_container redeployed at every chef-run if container exposes ports that are not published
In the same spirit as #444, I have a container that redeploys at every chef run because that container:
 - exposes ports 80 and 443 (indirectly through the base image, not from my own)
 - publishes port 80 (but not 443), which implies that we expose port 80

As a consequence, at each chef-run, the container redeploys because a "change" is detected. After instrumenting, I can read in the debug log:

```
[2015-09-21T14:54:02+02:00] DEBUG: DOCKER: exposed_ports - current:{"443/tcp"=>{}, "80/tcp"=>{}}: serialized:{"80/tcp"=>{}}:
```

I need to find a way so that when no expose ports are declared then all exposed ports from HostConfig are loaded in the serialized version.

jperville:
The handling of exposed_ports is tricky because some exposed_ports are injected from the image (EXPOSE instruction in Dockerfile). As such, we cannot rely on new_resource.exposed_ports and new_resource.port to declare which ports are actually be exposed.

The consequence is #446 where a container with multiple ports exposed from Dockerfile and at least one which is not bound (new_resource.port) is redeployed at every chef-run.

This PR fixes #446

### DIFF
--- a/libraries/helpers_container.rb
+++ b/libraries/helpers_container.rb
@@ -260,5 +260,11 @@ module DockerHelpers
       return true if (!new_resource.hostname.nil?) && (current_resource.hostname != new_resource.hostname)
       false
     end
+
+    def update_exposed_ports?
+      return false if new_resource.exposed_ports.nil?
+      return true if current_resource.exposed_ports != new_resource.exposed_ports
+      false
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


jperville:
I have another PR in my pipeline to fix a similar issue with volumes.

Issue #417, opened 2015-09-06
docker_image :pull fails against a private repo at hub.docker.com 


joegoggins:
On Ubuntu 14.04 or Centos 7.1 with Docker at 1.8.1, when I do this:

    docker_registry 'https://index.docker.io/v1/' do
      username 'the-username'
      email 'theemail@email.com'
      password 'the-password'
    end

    docker_image 'the-org/the-repo' do
      tag 'latest'
      action :pull
    end

I get this error:

    Docker::Error::ArgumentError
    ----------------------------
    Must have id, got: {"id"=>nil, :headers=>{}}

Oddly, if I do `docker login` with the same credentials and then `docker pull the-org/the-repo`, the image gets pulled down exactly as expected. I've tried a number of permutations of this and observed the same cookbook failure behavior. I'm guessing it has something to do with an upstream change in the Docker Registry api that the docker-api gem fails to handle gracefully. I noticed their is a new docker-api gem release, perhaps that might encapsulate this change...I don't know.

The use case I'm shooting for is a simple one: "Use docker cookbook to pull and deploy new private images from the main Docker registry", this issue is a blocker for this. 

Let me know if I can clarify the failure I've seeing or if anyone has suggestions for work arounds!

Thanks!

alexdeleon:
+1
Same issue here. Also on Centos 7.1.

someara:
This should be fixed in 1.0.22.

alexdeleon:
Hi @someara , I have tries with 1.0.22 but I still get the error. The only difference is that now the error contains the `X-Registry-Auth` header. 

```
RuntimeError
------------
Must have id, got: {"id"=>nil, :headers=>{"X-Registry-Auth"=>"XXXX"}}
```

joegoggins:
Not sure if it's helpful, but I vaguely recall this functionality working
fine on 1.0.14, but having the issue appear in 1.0.21.
I'm really looking forward to a fix on this, the particular route I take
toward a Docker infrastructure later today or tomorrow is highly contingent
on this working, let me know if there is anything I can do to test further
or provide more details.

Thanks,

-joe

On Thu, Sep 10, 2015 at 7:41 AM, Alexander De Leon <notifications@github.com

someara:
Ug.. I totally tested this manually here

https://github.com/bflad/chef-docker/blob/e626cffb80d84c370c6b546c4d941d54573c18fa/test/cookbooks/docker_test/recipes/image.rb#L197

[...]

PR #452, opened 2015-09-23
Fix registry auth and slightly refactor `docker pull`

pstengel:
This fixes registry authentication (hopefully for good), which is still broken in 1.0.32.

When you use `Docker.authenticate!`, it sets the credentials for *all* registry API calls. This means that we can't set credentials for multiple registries upfront -- each one you set simply overrides the last. It also creates problems like we saw in #417 when credentials for a private registry are used during a call to what's supposed to be a public registry.

Instead of using `Docker.authenticate!`, we instead parse `docker_registry`'s `serveraddress` as the real `docker pull` command does and store the credentials in the node's `run_state`. We then manually verify credentials using `Docker.connection`.

We also need to refactor `docker pull` emulation:

https://github.com/BWITS/docker-api/blob/9d4ba7df1a6b52bf4d4ff1ee04c160b06939717f/README.md

According to the above link, the proper way to emulate a `docker pull` is to use `Docker::Image.create`. This call accepts a Hash containing credentials for the image creation as the second argument. It also rids of all the workarounds in place that attempted to solve #417.

### DIFF
--- a/libraries/helpers_auth.rb
+++ b/libraries/helpers_auth.rb
@@ -1,0 +1,6 @@
+module DockerHelpers
+  # https://github.com/docker/docker/blob/4fcb9ac40ce33c4d6e08d5669af6be5e076e2574/registry/auth.go#L231
+  def parse_registry_host(val)
+    val.sub(%r{https?://}, '').split('/').first
+  end
+end
--- a/libraries/provider_docker_image.rb
+++ b/libraries/provider_docker_image.rb
@@ -70,25 +70,19 @@ class Chef
 
       def pull_image
         begin
-          cred_stash ||= Docker.creds
           retries ||= new_resource.api_retries
           api_timeouts
-          o = Docker::Image.get(image_identifier) if Docker::Image.exist?(image_identifier)
-          i = Docker::Image.create(
-            'fromImage' => new_resource.repo,
-            'tag' => new_resource.tag
-          )
-        rescue Docker::Error::ArgumentError => e
-          # Dirty hack around what seems to be a bug (feature?) in
-          # docker-api authentication header handling when pulling
-          # from the public registry with invalid creds are set
-          Docker.creds = {}
+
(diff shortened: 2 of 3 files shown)


chasebolt:
you removed the ability to constantly pull an image and only update the resource if a new image was pulled.

```
 o = Docker::Image.get(image_identifier) if Docker::Image.exist?(image_identifier)
```

registry_host should go into a parsing function as to not duplicate code.

foodcritic threw a few errors

```
FC001: Use strings in preference to symbols to access node attributes: ./libraries/provider_docker_image.rb:80
FC001: Use strings in preference to symbols to access node attributes: ./libraries/provider_docker_registry.rb:15
FC001: Use strings in preference to symbols to access node attributes: ./libraries/provider_docker_registry.rb:24
```

pstengel:
:+1: -- I'll commit fixes shortly

someara:
Thanks Paul! Looking forward to this one!

-s

pstengel:
Implemented @chasebolt's suggestions. Do you want a squash, or are the two commits okay?

someara:
Two commits are fine. I'll take a look at this in the morning. Thanks!
-s

someara:
I managed to get to this today!

I'm getting a converge failure right out the gate:
https://gist.github.com/someara/e93b3d20ab8d5630a606#file-gistfile1-txt-L166

You're requiring the new helper file, but never mixing in the methods...
I'd suggest namespacing it so you're not directly under DockerHelpers, then doing a

include DockerHelpers::Authentication

pstengel:
Blah. I tested the converge before I refactored into a helper, but not after. My fault. 

I'll try to get the fix in later tonight or first thing tomorrow. 

-Paul

someara:
no problem! this is why we have tests =)

-s

someara:
I'm going to just go ahead and fix this up for you... one sec

someara:
I manually fixed and squashed these as e1bbc68a90de4e8114b04e3e1e0f047e0119821f

someara:
Releases as 1.0.33

pstengel:
:+1:

PR #456, opened 2015-09-24
Fixed version detection of Docker version for rhel >= 7 to return 1.7…

jsparknz:
….1, so that the Docker daemon option is correctly selected on RHEL/CENTOS 7 and above.

### DIFF
--- a/libraries/helpers_service.rb
+++ b/libraries/helpers_service.rb
@@ -80,6 +80,7 @@ module DockerHelpers
       return '1.6.2' if node['platform'] == 'amazon'
       return '1.6.2' if node['platform'] == 'ubuntu' && node['platform_version'].to_f < 15.04
       return '1.6.2' if node['platform_family'] == 'rhel' && node['platform_version'].to_i < 7
+      return '1.7.1' if node['platform_family'] == 'rhel' && node['platform_version'].to_i >= 7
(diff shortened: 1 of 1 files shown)


chasebolt:
version 1.8.2 (which this originally returned) is perfectly acceptable for centos/rhel 7. i dont see why this change is needed.

jsparknz:
Thanks for the feedback. From what I can see, the latest version in the CentOS repositories is docker 1.7.1:

Available Packages
Name        : docker
Arch        : x86_64
Version     : 1.7.1
Release     : 115.el7
Size        : 7.0 M
Repo        : extras/7/x86_64
Summary     : Automates deployment of containerized applications
URL         : http://www.docker.com

Am I missing something here?

chasebolt:
this does not use repositories for installing docker. this downloads a binary from docker.io.

jsparknz:
Oh I see, my mistake sorry. We are installing the current version in the CentOS repositories. Would you be open to me modifying it to call the docker executable with the "-v" flag and parsing the output to get the real version number of the docker that is installed on the node?

someara:
Just set the version on the resource for now... using the package repos is
on the map

someara:
closing due to inactionability

Issue #458, opened 2015-09-26
docker_image :build_if_missing - Docker::Error::TimeoutError read timeout reached


kkeane:
This issue is similar to #403, but for the build_if_missing (and presumably build) action. I confirmed that the timeout works as expected for build_if_missing if the timeout is due to pulling a FROM image, but not when the timeout occurs during a RUN or other dockerfile entry.

In the following sample, if the image centos:latest was missing, the timeout occurred after approximately 2100 seconds (total chef runtime). If the image centos already existed, the timeout occurred after approximately 160 seconds (again total chef runtime).

    docker_image("redacted") do
      action [:build_if_missing]
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      declared_type :docker_image
      cookbook_name "redacted"
      tag "latest"
      source "redacted"
      force true
      read_timeout 3600
      write_timeout 3600
      repo "redacted"
      rm true
    end

Dockerfile:

FROM centos

COPY $LOCAL_DIR/*.rpm /tmp/

WORKDIR /tmp/

RUN \
   curl -SL https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm -o epel.rpm \
   && yum install -y epel.rpm \
   && rm epel.rpm \
   && yum install -y \
          <about 20 local RPMs with about 100 dependencies redacted>\
          supervisor \
   && yum clean all \
   && rm -rf /usr/share/man /etc/httpd/conf.d/ssl.conf

COPY supervisord.conf /etc/supervisord.d/supervisord.ini

COPY *.cfg /etc/redacted/

COPY entrypoint.sh /

RUN \
    chown root:root /entrypoint.sh \
 && chmod u+x /entrypoint.sh

CMD ["/usr/bin/supervisord", "-c", "/etc/supervisord.conf" ]
ENTRYPOINT ["/entrypoint.sh"]

WORKDIR redacted
EXPOSE 80 443

someara:
I can't reproduce this... can you submit a PR with a test recipe in it that should work, but does not?

https://github.com/bflad/chef-docker/tree/2549dd4386c24ee01387a656dc14611805daf979/test/cookbooks/docker_test/recipes

Thanks

-s

kkeane:
[...]

PR #461, opened 2015-09-30
Added test case for Issue #458 - timeout during docker_image :build_if_missing

kkeane:
As requested in the issue, here is the PR with the new test case.


someara:
I'll merge this when I have the bandwidth to make this pass.
Thanks!
-s

PR #471, opened 2015-10-01
Typo in Readme

ltouro:
Hostname should be declared as host_name on resource.


chasebolt:
it can be declared either way since an alias is defined. ``alias_method :hostname, :host_name``

ltouro:
Thanks for the info! :+1:

PR #473, opened 2015-10-02
fix for failed tls

### LINKED ISSUE #468: TLS verification broken
Recent fixes have revealed that the "service-tls" suites were actually connecting over a socket instead of over the network/TLS.

While fixing that, I noticed that the TLS verification is broken, at least while using the server certs for the client connection.

A temporary workaround is to disable TLS verification here:

https://github.com/bflad/chef-docker/blob/0a2f56449b42aa7a92d21095432688d1378431b5/libraries/provider_docker_service.rb#L26-L27

This is very dangerous, and it turns off TLS verification for the entire chef-client run.

Figuring out what exactly is going on here and fixing it is a prerequisite for #322 and precludes moving these resources into core.

chasebolt:
a lot of changes here. i did some cleanup in a couple of commits for clarity prior to the big tls changes.

i removed using the environment variables for the docker command since it doesnt match exact with the api. if a user had a cert and key in different locations it would have failed. there is now a docker_cmd to use when we need to shell out to docker cli. this takes care of any options we need to specify (tls and host atm).

the docker daemon uses multiple host commands, but the client only uses one so i created a parsing function for the connect_host. this should be expanded on later and should prioritize using the socket over tcp. for now though the function just returns the first element.

let me know how it tests out with test-kitchen.

resolves #468

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,10 +178,12 @@ section below for more information.
 ```ruby
 docker_service 'tls_test:2376' do
   host ["tcp://#{node['ipaddress']}:2376", 'unix:///var/run/docker.sock']
-  tlscacert '/path/to/ca.pem'
-  tlscert '/path/to/server.pem'
-  tlskey '/path/to/serverkey.pem'
-  tlsverify true
+  tls_verify true
+  tls_ca_cert '/path/to/ca.pem'
+  tls_server_cert '/path/to/server.pem'
+  tls_server_key '/path/to/serverkey.pem'
+  tls_client_cert '/path/to/client.pem'
+  tls_client_key '/path/to/clientkey.pem'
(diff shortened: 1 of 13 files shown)


chasebolt:
Also, I have only tested this with self-signed certs.

someara:
I merged this because its definitely a step in the right direction

However, test-kitchen explodes with an SSL verification error when converging service-tls

https://gist.github.com/someara/e008d2e8887420f012bc#file-gistfile1-txt-L7

-s

chasebolt:
Ok I'll take a look on Monday. Self signed certs work fine on my test
machine, I'll go over the test kitchen suite and see how it differs.

PR #487, opened 2015-10-06
Initial change to use 12.5 resource syntax

jkeiser:
- depend on compat_resource cookbook for backwards compatibility down to 12.0
- actions in resource instead of property
- subclass resources instead of using providers
- attribute -> property

There are more changes that can be made once this goes through; coercion, better defaults, nicer load_current_value among them.


someara:
I added some ChefSpec around a resource in the test recipe for the purposes of testing this against regressions... I found a world of unexpected behavior.

Let's figure out what's going on there first, then rebase this.

someara:
Whoopsies. I wrote the docker_image spec against the docker_container resource.
Fixed on master.

Working on a proper docker_container spec

someara:
docker_test::resources now has full ChefSpec coverage.

Hopefully you should be able to find errors via rspec that are causing the resources test-kitchen suites to fail on your branch.

-s

someara:
Merging another branch. Closing this! thanks!

PR #489, opened 2015-10-07
Update helpers_service.rb

agarat:
According to the [Docker reference] (https://docs.docker.com/reference/commandline/daemon/#storage-driver-options) there's no `=` between `--storage-opt` and its option.


chasebolt:
docker daemon options shows this in the help. i dont think this PR is needed as storage-opts works AFAIK.

```--storage-opt=[]                       Set storage driver options```

someara:
It looks like the actual behavior is to accept it both ways... let's call this a bug if the Docker documentation and keep this the way it is for consistency.

Thanks for the PR!

-s

PR #491, opened 2015-10-08
return best host for docker to connect to

### LINKED ISSUE #488: docker_wait_ready fails with bad certificate error
The docker_wait_ready block reuses the host options used to configure the docker service.
When binding the docker service to 0.0.0.0 this results in an error (that the tls certificate is not configured for that address).
While this could be fixed by adding 0.0.0.0 to the SAN fields, I don't think that makes sense.

I've worked around this by changing the order of the bindings in the service (so that the unix socket binding is in the 0 index of the array and the https://0.0.0.0 is in the second index of the array) but that won't help if someone chooses not to bind a socket address at all.

The most complete solution to this would probably:
1. In general, iterate on each of the bindings instead of copying the options as is
2. Try the unix socket first if configured
3. Second try any non 0.0.0.0 ip's configured,
4. Lastly, if bound to 0.0.0.0, try any public ip addresses on the machine, followed by any private ips, followed by localhost on the machine

chasebolt:
this could probably be done a lot better, but it works for now.

resolves #488

### DIFF
--- a/libraries/docker_service.rb
+++ b/libraries/docker_service.rb
@@ -112,19 +112,17 @@ class Chef
         def load_current_resource
           @current_resource = Chef::Resource::DockerService.new(name)
 
-          Docker.url = connect_host if connect_host
-
-          if connect_host =~ /^tcp:/ && tls_ca_cert
-            Docker.options = {
-              ssl_ca_file: tls_ca_cert,
-              client_cert: tls_client_cert,
-              client_key: tls_client_key,
-              scheme: 'https'
-            }
+          connect_opts = {}
+          if connect_host =~ /^tcp:/
+            connect_opts[:scheme] = 'https'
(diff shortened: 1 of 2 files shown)


someara:
Hello!

Sorry for the delay on looking at this... was busy working on the 2.0,0 release.

Can you update this for 2.0.0?

It should be as easy as sticking it in a helper function and calling it as a lazy default here:

https://github.com/bflad/chef-docker/blob/f2e346b31513ce7465157921e189a17c35d40229/libraries/docker_container.rb#L56

-s

chasebolt:
No worries Sean! I see a ton of changes in 2.0.0. I'll get this updated tomorrow.

chasebolt:
This is GTG :+1:

chasebolt:
hmm dont merge this yet, test-kitchen isnt happy.

chasebolt:
ok this is good to merge in now

PR #495, opened 2015-10-12
opts shouldnt be forced to true/false

### LINKED ISSUE #494: Disabling iptables in docker service
As I see [here](https://github.com/bflad/chef-docker/blob/f2e346b31513ce7465157921e189a17c35d40229/libraries/helpers_service.rb#L151) this code 
``` 
opts << '--iptables=true' if iptables
```
does nothing if iptables is set to false. 

This lead me to assume that --iptables=false is the default option, however, looking at the docker [docs](https://docs.docker.com/articles/networking/), it seems that this is not the case...? 
 
> Do your iptables allow this particular connection? Docker will never make changes to your system
> iptables rules if you set --iptables=false when the daemon starts. Otherwise the Docker server will
> append forwarding rules to the DOCKER filter chain.

am I missing something here guys?

chasebolt:
the boolean property should be passed through instead of forcing it to be true/false. originally i was only fixing iptables, but i decided to fix the remaining as well.

resolves #494

### DIFF
--- a/libraries/helpers_service.rb
+++ b/libraries/helpers_service.rb
@@ -148,7 +148,7 @@ module DockerHelpers
       opts << "--ip=#{ip}" if ip
       opts << "--ip-forward=#{ip_forward}" unless ip_forward.nil?
       opts << '--ip-masq=true' if ip_masq
-      opts << '--iptables=true' if iptables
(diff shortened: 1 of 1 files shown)


bflad:
:+1:

PR #496, opened 2015-10-12
Use with_retries and load_current_value in DockerImage

jkeiser:


someara:
Had to revert this... it was exploding test-kitchen.

https://gist.github.com/someara/f4de54614165a5681ea7

PS: The specs on this cookbook are only ChefSpec against the test recipes.

That's just enough to make sure that the resource collection contains the correct resources with the correct properties set. The specs aren't high enough resolution to catch this kind of thing, so actually performing the chef-client run is the only way to be sure.

-s

PR #505, opened 2015-10-13
changing container resource to default action :run

chasebolt:
action run has become very stable since we switched to using the new custom resource. i think run should become the new default so containers will be redeployed on resource changes.


someara:
pretty sure this is gonna require a trip through the test cookbook /
serverspecs to be mergable
-s

someara:
er, releasable

chasebolt:
passes test-kitchen :+1:

chasebolt:
looking into travis-ci complaints now

chasebolt:
all fixed up now

PR #521, opened 2015-10-15
daemon can accept multiple labels. updating README. string is valid

chasebolt:
adding label support to the daemon. also, occasionally docker inspect returns container labels unsorted and causes rspec to fail; implemented fix for this.


bflad:
LGTM

PR #528, opened 2015-10-21
dont enable https if we arent using tls

### LINKED ISSUE #526: SSL_connect error when running docker_service with host set
I'm trying to use the docker cookbook to setup a swarm. When creating the docker service, I set the host attribute as follows, per documentation: https://docs.docker.com/swarm/install-manual/

    docker_service 'default' do
      host 'tcp://0.0.0.0:2375'
      action [:create, :start]
    end

The converge bombs out with the following error:
    ================================================================================
    Error executing action `create` on resource 'docker_service_systemd[default]'
    ================================================================================

    Excon::Errors::SocketError
    --------------------------
    SSL_connect returned=1 errno=0 state=SSLv2/v3 read server hello A: unknown protocol (OpenSSL::SSL::SSLError)

However, if I then log into the box using Kitchen, and run the info command with the host specified, the service was created and appears to be running:

    [root@swarm-manager-centos71 ~]# docker -H tcp://0.0.0.0:2375 info
    Containers: 0
    Images: 0
    Storage Driver: devicemapper
    Pool Name: docker-253:0-135301711-pool
    Pool Blocksize: 65.54 kB
    ....

I can also issue other docker commands like pull, run, etc.

Some details:
OS: CentOS Linux 7 (Core) - From vagrant box: boxcutter/centos71
Docker Cookbook Version: 2.1.2
ChefDK: 0.8.0

chasebolt:
resolves #526

### DIFF
--- a/libraries/docker_service.rb
+++ b/libraries/docker_service.rb
@@ -115,7 +115,7 @@ class Chef
 
           connect_opts = {}
           if connect_host =~ /^tcp:/
-            connect_opts[:scheme] = 'https'
+            connect_opts[:scheme] = 'https' if tls || !tls_verify.nil?
             connect_opts[:ssl_ca_file] = tls_ca_cert if tls_ca_cert
             connect_opts[:client_cert] = tls_client_cert if tls_client_cert
             connect_opts[:client_key] = tls_client_key if tls_client_key


someara:
excellent!

PR #541, opened 2015-11-03
fix obsolate "fi" at the end of post-start script

epcim:
Found upstart service to fail on start due an lost "fi" from the "if" statement at the end of the script.


jperville:
Well spotted @epcim

PR #546, opened 2015-11-05
Chef fails if docker v1.9.x service already started

scalp42:
See issue https://github.com/someara/chef-docker/issues/545

This will fix the return code for the Upstart job if already running.

Please test with Docker v1.8.3 as I don't have access to a box with it immediately.

cc @someara for review :smile:

PS: Your work is much appreciated as always #hugops


chasebolt:
LGTM. We should add a TODO for refactoring this Upstart script.

scalp42:
+1 for refactoring the logic (just wanted to patch for now).

PR #565, opened 2015-11-25
Add support for --exec-opt to docker daemon

gjacquet:
Add support for --exec-opt to docker daemon.
This seems to be needed to workaround docker/docker#17653 on Centos 7.


chasebolt:
LGTM

PR #566, opened 2015-12-01
Rename cluster-opts to cluster-opt

smgt:
Noticed that the naming of the `--cluster-store-opt` was wrong, a `s` to much.

From `docker daemon --help`
```
--cluster-store-opt=map[]            Set cluster store options
```


chasebolt:
Thanks for catching my plural issue!

I would keep an `=` on the opt=value for code consistency. Other than that LGTM.

smgt:
Fixed the `=`! And you should not blame yourself, 6bc5c7fd3b9cc68ccdc6840207875e4c6f70bb92. It was my plural issue. :smile_cat:

chasebolt:
Ahh I remember this now. I had a PR ready to go for this but you beat me to
it ;)

LGTM 👍

PR #568, opened 2015-12-04
Add a network resource

smgt:
I just wanted to put this in. It works but the vendored version of swipely/docker-api is a special version with network support. There is a pull request ( swipely/docker-api#335 ) that is waiting to get merged. Please have a look and consider it. It includes tests.
It depends on the consul cookbook, you can't add networks when docker isn't started in cluster mode and as far as I can see BoltDB isn't fully implemented to work in docker.


chasebolt:
One thing to throw in before you get deep. I don't think we want to depend on consul. Since Docker just needs a KV store, that should be up to the user's preference.

chasebolt:
I am not a fan of git submodules, not sure how @someara feels on this topic though.

I do like the addition of the docker_network resource :+1:

smgt:
> One thing to throw in before you get deep. I don't think we want to depend on consul. Since Docker just needs a KV store, that should be up to the user's preference.

The dependency is in the `test_docker` cookbook, not in the main cookbook. I think it is a OK trade off to be able to test the network resource. I really, really wanted to use BoltDb so we didn't need to add consul to the `test_docker` cookbook. But since docker don't register boltdb as a kv store for discovery it will be hard to use it.

https://github.com/smgt/docker/blob/0e4fc8aa7e522e9b03f3817d8e8028186762ece4/test/cookbooks/docker_test/metadata.rb#L10

> I am not a fan of git submodules, not sure how @someara feels on this topic though.

I hate submodules, I just wen't with it since it is easier when updating several git repositories at the same time. I'm planning to remove the submodule when swipely/docker-api#335 is in master.

> I do like the addition of the docker_network resource :+1:

This is a work in progress, I just wanted to share so that not to many jump in and write a new network resource. I would love feedback on the resource and tests since this is one of my first cookbooks. :smile_cat:

someara:
Hi Simon!

Looking good so far... I'll take a closer look on Monday.

Thanks for the heads up! I was thinking about implementing this ;)

-s

gionn:
Hi, I am interested to have network support into the cookbook, someone is still working on it?

smgt:
@gionn Yes, the block is swipely/docker-api#335. The support needs to get merged into master for this PR.

smgt:
Ok, so swipley/docker-api#335 got merged into master an released as 1.25.0. I've vendored the new version of `docker-api` and rebased on master. ping @someara @chasebolt

someara:
Hi!
First off,  Thank you! This is awesome.

I manually rebased and merged this in along with some fixes... I wanted to get it in before it diverged too far. I switched to using etcd as cluster_store because the consul cookbook has too many deps and one was throwing weird errors. https://gist.github.com/someara/64fe20952431409a6796#file-gistfile1-txt-L80

I'm dark launching it by releasing it in patch version and not adding it to the README.

There are a few issues remaining...

First, we need to get some inspec tests in there for kitchen verify...
Second, I noticed that the test recipe fires 4 resources on the second test-kitchen run.

Once those are resolved, we can get this into the README and do a feature version bump =)

-s

PR #574, opened 2015-12-10
updating docker to 1.9.1

chasebolt:


chasebolt:
pull this PR last since it renames rspec directories; otherwise it will run into a conflict.

PR #578, opened 2015-12-11
Update docker_container library documentation on timeouts

jsm:


chasebolt:
the read/write timeout on the container resource is actually valid and is inherited from docker_base.

you are correct in noticing there is an undocumented `timeout` property. when a container is not detached, this is the length of time to wait for an attached container to return. it would be great if you can update your commit to reflect that in the doc.

jsm:
Hah, guess all I'm doing now is just adding the timeout documentation then.

chasebolt:
:smile: LGTM

PR #585, opened 2015-12-12
Typo fix: it should be pidfile property

artem-sidorenko:
Otherwise you might get `NoMethodError: undefined method `pid_file' for Chef::Resource::Execute`


someara:
released in v2.3.10

PR #587, opened 2015-12-15
Allow distributed gems to be activated normally via Ruby

jkeiser:
Handles these cases better:
1. If docker-api is not loaded or installed on the system, the cookbook’s version will be used.
2. If docker-api is not loaded, and docker-api 1.30 is installed on the system, that will be loaded instead.
3. If docker-api is not loaded, and docker-api 1.23 is installed on the system, the cookbook's version (1.24) will be used.
4. If someone else loaded docker-api 1.23 already, we fail to load the cookbook since it's not an acceptable version.
5. If someone else loaded docker-api 1.30 already, we don't load anything because it is an acceptable version.


someara:
rubocop -a =)

PR #590, opened 2015-12-17
parse binds to match docker cli parsing of -v

### LINKED ISSUE #493: docker_container :create fails with container-only binds property element
I reproduced this with the head of the git repository as of right now (10/11, around noon PST), haven't tried with a released version, so this could potentially be a transient problem. I stumbled across this trying to reproduce #492 . I am using docker 1.7.1 (docker API version 1.19), on CentOS 6.7.

If you specify a container volume in the binds array without the corresponding host volume, creating the container will fail, claiming that the container already exists, even though it actually doesn't.

The semantics of specifying only a container volume without a host volume is that the volume should be placed under /var/lib/docker/volumes/xxxxx/_data (the same semantics as the VOLUME entry in a Dockerfile).

Preparation: make sure that the container "bindtest_chef" is deleted.

Details:

The following resource will always fail with the message "Conflict. The name "bindtest_chef" is already in use by container a7d892f46981. You have to delete (or rename) that container to be able to reuse that name."

    docker_container "bindtest_chef" do
      action :create
      repo "centos"
      binds [ '/home', '/tmp/bindtest:/tmp' ]
    end

The following resource will succeed (the only difference is the first entry in the binds array):

     docker_container "bindtest_chef" do
      action :create
      repo "centos"
      binds [ '/home:/home', '/tmp/bindtest:/tmp' ]
    end

chasebolt:
with great pleasure i bring you this fantastic, over the top, (some say) majestic, PR that resolves #493

### DIFF
--- a/README.md
+++ b/README.md
@@ -1019,10 +1019,11 @@ Most `docker_container` properties are the `snake_case` version of the
 - `command` - The command to run when starting the container.
 - `autoremove` - Boolean - Automatically delete a container when it's
   command exits. Defaults to `false`.
-- `binds` - An array of `:` separated paths to bind mount from the
-  host into the container in the form
-  `['/host-bits:/container-bits', '/more-host-bits:/more-container-bits']`.
-  Defaults to `nil`.
+- `binds` - An array of volume bindings for this container. Each volume binding
+   is a string in one of these forms:
+    `container_path` to create a new volume for the container.
+    `host_path:container_path` to bind-mount a host path into the container.
+    `host_path:container_path:ro` to make the bind-mount read-only inside the container.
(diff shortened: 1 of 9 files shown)


someara:
I merged this thinking I could flip the logic around quickly to make volumes the canonical property.
Ended up backing it out for now.

Will take another crack at it tomorrow

chasebolt:
Hah yea it wasn't so easy. I think it is mainly done in my branch that I haven't pushed up yet. I don't get back from Philippines until Jan 10th though.

someara:
No problem!

I'm going to be in Seattle next week anyway, so I'm going to be a little
slow with the maintainering anyway.

-s

PR #599, opened 2016-01-10
Fix #598. return value check for docker-wait-ready with systemd

### LINKED ISSUE #598: return value check for docker-wait-ready with systemd
**VERSION**
2.4.0
Commit 052f747

**DESCRIPTION**
The script at `templates/default/systemd/docker-wait-ready.erb` is checking for a wrong iterator number to return the failure case.

**REPRODUCE**
The easiest way to test is modifiying docker-wait-ready.erb and introducing at grep ^FAIL-CONTAINER to force failure

odacremolbap:
**VERSION**
2.4.0
Commit 052f747e8c0f69b177137d698226228f2885d093

**REPRODUCE**
( Edited after @bflad 's comment)
Use this modified docker-wait-ready.erb file to test failure
```
#!/usr/bin/env bash
i=0
while [ $i -lt <%= @service_timeout * 2 %> ]; do
  <%= @docker_cmd %> ps | head -n 1 | grep ^failureCONTAINER > /dev/null 2>&1
  [ $? -eq 0 ] && break
  ((i++))
  sleep 0.5
done
[ $i -eq <%= @service_timeout * 2 %> ] && exit 1
exit 0
```

### DIFF
--- a/libraries/docker_service_manager_systemd.rb
+++ b/libraries/docker_service_manager_systemd.rb
@@ -16,6 +16,8 @@ module DockerCookbook
       node['platform_version'].to_f >= 15.04
     end
 
+    property :service_timeout, Integer, default: 20
+
     action :start do
       # Needed for Debian / Ubuntu
       directory '/usr/libexec' do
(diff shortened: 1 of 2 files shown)


bflad:
This is definitely an improvement. There however is still bad logic for the final iteration. If the container does actually start, it'll still return `exit 1` due to the `i+1` being above the `break`

odacremolbap:
@bflad , you are right. Fixed changing the addition after the break.
Thanks a lot.

PR #607, opened 2016-01-19
volumes is the new binds

### LINKED ISSUE #493: docker_container :create fails with container-only binds property element

chasebolt:
parsing and coercing volumes to now set the proper binds and volumes config in docker. we were just being willy nilly about this before and passing it to docker incorrectly.

binds is no more, volumes is the new hipster way to do it.

resolves #493

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ docker_container 'my_nginx' do
   host_name 'www'
   domain_name 'computers.biz'
   env 'FOO=bar'
-  binds [ '/some/local/files/:/etc/nginx/conf.d' ]
+  volumes [ '/some/local/files/:/etc/nginx/conf.d' ]
 end
 ```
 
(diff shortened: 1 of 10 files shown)


someara:
Thanks Chase!

I squashed the commits into 9e0d173a8048d5117413c36f8a11d930d82bf172
Released as 2.4.2

PR #611, opened 2016-01-20
Use Gem::Version to handle semantic verisoning and be compatible for ≥ 1.10

Soulou:
The current 'master' code is not working for docker >= 1.10

Explanation:

``` ruby
"1.10".to_f < 1.8
```

It results in running `docker -d` instead of `docker daemon`

This patch fix that by using `Gem::Version` class which handle pretty well versions, and even, pre-version.

### DIFF
--- a/libraries/helpers_service.rb
+++ b/libraries/helpers_service.rb
@@ -92,7 +92,7 @@ module DockerCookbook
       end
 
       def docker_daemon_arg
-        if docker_major_version.to_f < 1.8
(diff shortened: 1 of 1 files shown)


someara:
Hello!

We need to check the version on the docker binary, not the version of the docker-api gem.
-s

Soulou:
True, that exactly what it is doing.

The check is using `Gem::Version` as a tool to understand semantic versioning, it is not looking at the `docker-api` gem at all.

Soulou:
Great thanks!

Issue #428, opened 2015-09-16
docker_container and broken volume definitions


jperville:
In the same spirit as #427, I found some differences in the `docker inspect` output when trying to declare a container which volumes, which resulted in bind mount being unusable with that container.

The test container is a container based upon the busybox:ubuntu-14.04 image which uses a bind-mount to the local filesystem (eg. `docker run ... -v /docker-images:/images/ro ...`) and at the same time bind mount from a data container using `volumes_from`.

The container is declared like this (simplified version):
```ruby
docker_container 'my-test-app' do
  repo 'busybox'
  tag 'ubuntu-14.04'
  volumes_from 'my-test-data'
  volume '/docker-images:/images:ro'
  port ['7080:80', '7022:22'] # bug #427
  command '/bin/top'
  ...
end
```

The following if the diff of `docker inspect` between:
 * the container as generated by this cookbook
 * the container recreated manually with the docker client.

TL;DR: the main difference is the `Binds` and `.Mount` entries, which are broken on the cookbook version. The diff also displays bug #427 in the `PortBindings` section and another bug of unknown consequences related to hostname/domainname parsing that will report next.

```diff
--- inspect-cookbook.json       2015-09-16 12:47:18.568194819 +0000
+++ inspect-manual.json 2015-09-16 13:01:22.084194819 +0000
@@ -63,7 +63,9 @@
     "AppArmorProfile": "",
     "ExecIDs": null,
     "HostConfig": {
-        "Binds": [],
+        "Binds": [
+            "/docker-images:/images:ro"
+        ],
         "ContainerIDFile": "",
         "LxcConf": null,
         "Memory": 67108864,
@@ -78,15 +80,15 @@
         "MemorySwappiness": null,
         "Privileged": false,
         "PortBindings": {
-            "22": [
+            "22/tcp": [
                 {
-                    "HostIp": "0.0.0.0",
+                    "HostIp": "",
                     "HostPort": "7022"
                 }
             ],
-            "80": [
+            "80/tcp": [
                 {
-                    "HostIp": "0.0.0.0",
+                    "HostIp": "",
                     "HostPort": "7080"
                 }
             ]
@@ -143,24 +145,22 @@
             "RW": true
         },
         {
-            "Name": "461673042e45a1defcb709c8971bf2eb00986553bb0e06a726d3cfadbb55d89e",
[...]

PR #620, opened 2016-01-27
README volumes -> binds

ptqa:
I believe that is how volumes/binds should be described in README. Because currently it's misleading and people have problems because of that (see https://github.com/chef-cookbooks/docker/issues/428).

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ docker_container 'my_nginx' do
   host_name 'www'
   domain_name 'computers.biz'
   env 'FOO=bar'
-  volumes [ '/some/local/files/:/etc/nginx/conf.d' ]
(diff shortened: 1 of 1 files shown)


chasebolt:
This has been discussed and is a difference between the API and the CLI. API talks about binds while CLI talks about volumes (-v). We settled on calling them volumes as people work mostly with CLI and less with the API.

ptqa:
@chasebolt I'm not sure what do you mean. This PR is not about naming resource attributes in this cookbook. It's about keeping documentation up to date with code.

chasebolt:
Volumes is the property to use. We have deprecated binds (which is an alias to volumes now).

ptqa:
@chasebolt ah, I see. I probably use old version of cookbook.

PR #624, opened 2016-02-01
Add host back to docker network

smgt:
It seems that `host` property got removed from `Docker::Network` ( chef-cookbooks/docker@b743e8b609c9fd809399e19d2f4e164a24226e22 ). This makes it impossible to make a connection to the docker daemon ( https://github.com/chef-cookbooks/docker/blob/master/libraries/helpers_base.rb#L50 ). This PR will add `host` back so `docker_network` starts working again.

PS. I have tests for docker network on their way. DS.


smgt:
TY!

ping @qzio

Issue #597, opened 2016-01-05
Container being redeployed at every chef-run


jperville:
Currently using version 2.4.0 of this cookbook.

My global docker_service resource is configured to use `log_driver 'fluentd'` and `log_opts "fluentd-address=127.0.0.1:24224"`.
I have a docker_container that I explicitly configured to use `log_driver 'json-file'` and explicitly specifies `log_opts Hash.new`.

At each chef-run, that docker_container is redeployed with the following trace:

```
    * docker_container[registrator] action run
      - stopping registrator 
      - deleting registrator
      - update registrator
      -   set exposed_ports to {} (was nil)
      -   set log_opts      to {} (was {"fluentd-address"=>"127.0.0.1:24224"})
      - starting registrator
```

Please note that here is how the docker-inspect looks like:
```sh
root@staging4-compute:~# docker inspect -f '{{.HostConfig.LogConfig}}' registrator
{json-file map[fluentd-address:127.0.0.1:24224]}
```
Also note that the image I am using (`gliderlabs/registrator:v6`) is not exposing any port.

I believe that the redeploy is due to incorrect attribute merging for the log_opts attribute; docker blindly reuses the docker_service value (even if it is incorrect for the container's log-driver and if docker won't start the container if I supply incorrect values for the log_opts) and I have no way to tell the cookbook explicitly to use `Hash.new` as the log_opts (the cookbook thinks that what I supplied is the default value and won't override the docker-supplied incorrect value).

jperville:
Hi @someara , I'm still having the issue with 2.4.3 (after fixing a bug with immutable array for which I'm about to submit a PR). Please consider reopening.

```
    * docker_container[registrator] action run
      - stopping registrator 
      - deleting registrator
      - update registrator
      -   set exposed_ports to {} (was nil)
      -   set log_opts      to {} (was {"fluentd-address"=>"127.0.0.1:24224"})
      - starting registrator
```

scalp42:
@someara correct I'm seeing containers being redeployed as well here:

```
  * docker_container[mysql] action create
    - stopping mysql
    - deleting mysql
    - update mysql
    -   set volumes to nil (was {"/var/lib/mysql"=>{}})
  * docker_container[mysql] action run
    - deleting mysql
    - update mysql
    -   set volumes to nil (was {"/var/lib/mysql"=>{}})
    - starting mysql
```

Recipe pretty straightforward:

```ruby
docker_image 'mysql' do
[...]

PR #631, opened 2016-02-05
Set default docker_container.exposed_port to empty Hash

jperville:
Isolated this bit from issue #597 and proposing a PR to fix the 'exposed_ports' case.

Currently using version 2.4.11 of this cookbook.

I have a docker_container on which redeploys at each chef-run. This container is using the `gliderlabs/registrator:v6` image, an image which does NOT expose any port all all, as shown by the following `docker inspect` commands:

```sh
# This image exposes no ports
$ docker inspect -f '{{ .Config.ExposedPorts }}' gliderlabs/registrator:v6
map[]
# Compare with a typical image which exposes ports
$ docker inspect -f '{{ .Config.ExposedPorts }}' phusion/passenger-ruby22:0.9.17
map[443/tcp:{} 80/tcp:{}]
```

At each chef-run, that docker_container is redeployed with the following trace:

    * docker_container[registrator] action run
      - stopping registrator 
      - deleting registrator
      - update registrator
      -   set exposed_ports to {} (was nil)
      - starting registrator

I believe that the redeploy is due to an incorrect default value for the docker_container exposed_port attribute. Since the API does not list the attribute in its response, the default value (`nil`) from the docker_container resource is assumed for the current resource while the value on the new resource is an empty Hash. The values differ, hence redeployment.

This PR sets the default value for exposed ports to an empty Hash. I ran out of time to write a dedicated test but I can confirm that the fix works on our servers and that the `resources-191-ubuntu-1504` kitchen suite still passes.

### DIFF
--- a/libraries/docker_container.rb
+++ b/libraries/docker_container.rb
@@ -49,7 +49,7 @@ module DockerCookbook
     property :entrypoint, ShellCommand
     property :env, UnorderedArrayType, default: []
     property :extra_hosts, NonEmptyArray
-    property :exposed_ports, PartialHashType
+    property :exposed_ports, PartialHashType, default: {}
     property :force, Boolean, desired_state: false
     property :host, [String, nil], default: lazy { default_host }, desired_state: false
     property :hostname, String


jperville:
For information, the travis failure is unrelated to this PR.

chasebolt:
LGTM

Issue #645, opened 2016-02-16
docker_network aux_address property broken


someara:
The aux_address property on the docker_network doesn't seem to be working.

https://github.com/chef-cookbooks/docker/blob/8f77c66b180e9bd998db975fe498620899b7ef3e/test/cookbooks/docker_test/recipes/network.rb#L76

Doing it by hand lets me figure out what the inspec should look like.

https://github.com/chef-cookbooks/docker/blob/f9b559c0e92f0bef1e9db5ba409bfa559b203d66/test/integration/network/serverspec/assert_functioning_spec.rb#L69-L73

-s

someara:
Paging @gionn

smgt:
Made a PR, kinda silly error. https://github.com/chef-cookbooks/docker/pull/646/files#diff-85ca069b2156679441b35497a4782412R38

someara:
Fixed this up with property coersion and converge_if_changed.
Released in 2.5.0

smgt:
Nice work!

PR #646, opened 2016-02-16
Fix aux error in docker network

smgt:
This will also add a couple of tests. This will create test duplicates. Issue #645

### DIFF
--- a/spec/docker_test/network_spec.rb
+++ b/spec/docker_test/network_spec.rb
@@ -59,4 +59,61 @@ describe 'docker_test::network' do
       )
     end
   end
+
+  context 'testing to connect container to network' do
+    it 'created a container' do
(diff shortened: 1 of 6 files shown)


spec/docker_test/network_spec.rb
```diff
@@ -59,4 +59,61 @@
...
+    it 'connects container to network' do
+      expect(chef_run).to connect_docker_network('test-network-connect').with(
+        container: 'network-container'
+      )
+    end
+  end
+
+  context 'testing ip range' do
+    it 'should set a ip range' do
+      expect(chef_run).to create_docker_netwrok('test-network-ip-range').with(
```
tacchino: Typo here

someara: Yep. On it...
I'm going to spend the day getting the network and volume bits polished.

PR #659, opened 2016-02-24
Removes vendored gems in favor of chef_gem install

kri5:
This should fix #649.

PR #663, opened 2016-02-26
fix serverspec for new gem install method

chasebolt:
this fixes #659 issues with serverspec failing to install the gem. relying on bundler to install the docker gem for serverspec. _autoload will try to load the gem and if it fails, we will then trigger chef to install it.

removing version pinning for excon and json; i dont see a need for this now that we are using chef_gem to handle the install.


someara:
manually merged in as 2de6d0cd0c0657c9396e0a87506a6acef40862f1

chasebolt:
Robbing me of my stats!! haha 😉

Awesome glad we have moved away from vendoring gems. Can now apply this new method to other cookbooks 🍻

someara:
nah, I faked the author in the commit =)

Issue #679, opened 2016-03-09
Redeploy fails when container is in 'restarting' state


CVTJNII:
During convergence there appears to be a race caused by containers in a restarting state.  v2.5 of the cookbook has been intermittently throwing the following error when attempting to redeploy a container:

```
             ================================================================================
             Error executing action `redeploy` on resource 'docker_container[InfluxDB]'
             ================================================================================
             
             Docker::Error::ConflictError
             ----------------------------
             Conflict, You cannot remove a running container. Stop the container before attempting removal or use -f
             
             Cookbook Trace:
             ---------------
             /tmp/kitchen/cache/cookbooks/docker/libraries/docker_container.rb:366:in `block (3 levels) in <class:DockerContainer>'
             /tmp/kitchen/cache/cookbooks/docker/libraries/helpers_base.rb:66:in `with_retries'
             /tmp/kitchen/cache/cookbooks/docker/libraries/docker_container.rb:366:in `block (2 levels) in <class:DockerContainer>'
             /tmp/kitchen/cache/cookbooks/docker/libraries/docker_container.rb:365:in `block in <class:DockerContainer>'
             /tmp/kitchen/cache/cookbooks/compat_resource/files/lib/chef_compat/copied_from_chef/chef/provider.rb:122:in `instance_eval'
             /tmp/kitchen/cache/cookbooks/compat_resource/files/lib/chef_compat/copied_from_chef/chef/provider.rb:122:in `compile_and_converge_action'
             /tmp/kitchen/cache/cookbooks/docker/libraries/docker_container.rb:159:in `call_action'
             /tmp/kitchen/cache/cookbooks/docker/libraries/docker_container.rb:356:in `block in <class:DockerContainer>'
             /tmp/kitchen/cache/cookbooks/compat_resource/files/lib/chef_compat/copied_from_chef/chef/provider.rb:122:in `instance_eval'
             /tmp/kitchen/cache/cookbooks/compat_resource/files/lib/chef_compat/copied_from_chef/chef/provider.rb:122:in `compile_and_converge_action'
             /tmp/kitchen/cache/cookbooks/monitoring_infra/recipes/influxdb.rb:91:in `block (2 levels) in from_file'
```

Dropping in some debug prints it looks like the container fails to stop as it is restarting:

```
[...]

Issue #510, opened 2015-10-13
docker_container polling operations are not always desirable. 


someara:
It turns out that the fix for #498 causes test-kitchen runs to go from a few minutes to over 20.

``` 
-----> Kitchen is finished. (23m40.86s)
zlib(finalizer): the stream was freed prematurely.

real    23m41.562s
user    0m9.277
sys     0m2.473s
laptop:~/src/opscode-cookbooks/docker$
```

We should probably add a property on the resource to make this optional while still letting redeploy be safe

-s

chasebolt:
Do you know what was hanging it up? Containers traversing to the running
state must take a lot longer than expected...

someara:
That's pretty much it. I was also surprised. Apparently the old behavior
was just "fork and forget"
-s

chasebolt:
What actually could be happening is a container go from created  > running
> completed faster than the wait method checks. This would cause running to
be false (since it already completed) and the wait to do a full timeout
cycle of 20 seconds.

We need to check one of the timestamp fields as well to catch this case.

jperville:
Yes, the polling for container running makes the kitchen run very slow (especially if the container crashes for another reason, then we have to wait 20 seconds at least per container)... We could probably reduce the "sleep 1" to a shorter delay to make this more reactive.

CVTJNII:
I don't believe the solution in #511 was the correct one, and it may have masked some other problems.  While working on #679 I modified this logic back to the default and I noticed that the full timeout was being hit, even though the container was in the desired state.  The timeout appears to be due to the fact that the container object is not being updated in the state() method.  As such wait_running_state() will block for the full timeout unless the container is already in the desired state as changes are not registered.  My current assumption is that multi-operation actions the calls to individual actions will cause load_current_value to refresh the container object.

I have a patch in 06ca5d6 which I believe resolves this and #679 completely which I'm still testing.  A PR may be delayed until my company gets a CCLA in place.

chasebolt:
Do you have a working test case for this that you can gist?

CVTJNII:
@chasebolt What kind of a tests are you looking for?  I figured this out with prints and 'watch docker ps' running in another window, so I don't have an automated test case.  I do have a test cookbook where redeploy takes 20s without the patch to state() but that's no good for automation.

I considered making wait_running_state() throw an exception on timeout, as something is probably wrong if the container doesn't reach the desired state in 20s and that would cause test failures, but I wasn't sure if others would agree.

[...]

PR #683, opened 2016-03-17
Resolve restarting container races

CVTJNII:
This PR:
- Ensures that wait_running_state uses the current state of the container, not a stored value, as discussed in #510 
- Throws an exception if wait_running_state times out.  This is used to allow retries for the second race mentioned in #679 and should cause test errors if the operation is timing out.
- Retries the stop operation if the container fails to stop in wait_running_state.  This resolves a difficult to reproduce corner case race mentioned in #679.


chasebolt:
looks good! you'll need to remove the version bump.

CVTJNII:
"you'll need to remove the version bump."

Done.

PR #728, opened 2016-07-04
Add extra options to daemon

willejs:
### Description

I am using this cookbook with flannel. Flannel creates a docker bridge, and exposes the docker bip subnet through an env file. I can override the systemd environmentFile option through using the docker-instance.d/ conf file easily.

However, to pass the environment variable in the env file to the docker daemon, i need to evaluate the var at runtime, by setting something like this ```--bip=${FLANNEL_SUBNET}``` at docker daemon runtime.

Therefore I need to set the bip property of the docker_service provider to ```${FLANNEL_SUBNET}```, however it fails IP address validation against the regex in the provider, understandably.

Therefore, we should be able to pass arguments as strings to avoid validation in these edgecases.

### Issues Resolved
None.

### Check List
- [x] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,8 @@
 source 'https://rubygems.org'
 
+# is needed until Chef DK has Ruby version >= 2.2.2
+gem 'activesupport', '=4.2.6'
+
 group :docker do
   gem 'docker-api', '= 1.28.0'
 end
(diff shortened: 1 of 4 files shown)


willejs:
@someara @thommay could you review please?

willejs:
@chasebolt Can you review please?

PR #730, opened 2016-07-14
Add uts_mode support for docker_container provider

strangeman:
### Description

Add uts_mode support for docker_container provider. 

https://docs.docker.com/engine/reference/run/#/uts-settings-uts here is option description.

> UTS settings (--uts)
> --uts=""  : Set the UTS namespace mode for the container,
>        'host': use the host's UTS namespace inside the container
> The UTS namespace is for setting the hostname and the domain that is visible to running processes in that namespace. By default, all containers, including those with --network=host, have their own UTS namespace. The host setting will result in the container using the same UTS namespace as the host. Note that --hostname is invalid in host UTS mode.
>You may wish to share the UTS namespace with the host if you would like the hostname of the container to change as the hostname of the host changes. A more advanced use case would be changing the host’s hostname from a container.

### Check List
- [ ] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ group :kitchen_common do
   gem 'test-kitchen', '~> 1.5.0'
   gem 'kitchen-sync'
   gem 'kitchen-inspec'
+  gem 'activesupport', '< 5.0.0'
 end
 
 group :kitchen_vagrant do
--- a/README.md
+++ b/README.md
@@ -1093,6 +1093,9 @@ Most `docker_container` properties are the `snake_case` version of the `CamelCas
 - `tls_client_cert` - Path to TLS certificate file for docker cli. Defaults to ENV['DOCKER_CERT_PATH'] if set
 - `tls_client_key` - Path to TLS key file for docker cli. Defaults to ENV['DOCKER_CERT_PATH'] if set
 - `userns_mode` - Modify the user namespace mode - Defaults to `nil`, example option: `host`
+- `pid_mode` - Set the PID (Process) Namespace mode for the container. `host`: use the host's PID namespace inside the container. 
+- `ipc_mode` - Set the IPC mode for the container - Defaults to `nil`, example option: `host`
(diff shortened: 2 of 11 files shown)


strangeman:
I had some troubles with integration tests: docker won't installed to old Debian, Centos, and Ubuntu:

```
       [2016-07-14T02:41:51+00:00] ERROR: docker_installation_package[default] (docker_test::installation_package line 1) had an error: Mixlib::ShellOut::ShellCommandFailed: apt_package[docker-engine] (/tmp/kitchen/cache/cookbooks/docker/libraries/docker_installation_package.rb line 17) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '100'
       ---- Begin output of apt-get -q -y install docker-engine=1.7.1-0~wheezy ----
       STDOUT: Reading package lists...
       Building dependency tree...
       Reading state information...
       Some packages could not be installed. This may mean that you have
       requested an impossible situation or if you are using the unstable
       distribution that some required packages have not yet been created
       or been moved out of Incoming.
       The following information may help to resolve the situation:
       
       The following packages have unmet dependencies:
        docker-engine : Depends: init-system-helpers (>= 1.18~) but it is not installable
                 Recommends: aufs-tools but it is not going to be installed
                 Recommends: cgroupfs-mount but it is not installable or
                             cgroup-lite but it is not installable
                 Recommends: git but it is not going to be installed
       STDERR: E: Unable to correct problems, you have held broken packages.
       ---- End output of apt-get -q -y install docker-engine=1.7.1-0~wheezy ----
       Ran apt-get -q -y install docker-engine=1.7.1-0~wheezy returned 100
       [2016-07-14T02:41:53+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
-----> Destroying <installation-package-171-debian-710>...
       ==> default: Forcing shutdown of VM...
       ==> default: Destroying VM and associated drives...
       Vagrant instance <installation-package-171-debian-710> destroyed.
```

But it not seems related to current functionality.

chasebolt:
released in v2.9.3

strangeman:
Thank you, Chase! :)

PR #734, opened 2016-07-27
Don't send command propery to docker API unless explicitly specified …

### LINKED ISSUE #673: Default Docker CMD not updated on image update
I'm using this cookbook to run a container that specifies a default CMD in the dockerfile.
For example, my Dockerfile has the line

```
CMD /usr/src/run.sh
```

If I run the container like:

```ruby
docker_image "my_image" do
  tag 'latest'
  action :pull
  notifies :redeploy, 'docker_container[my_container]', :delayed
end

docker_container "my_container" do
  repo "my_image"
  tag 'latest'
  port "8080:8080"
  restart_policy 'always'
end
```

This runs fine.
If I push a version of the image that changes the default CMD, e.g. update the Dockerfile with

```
CMD /usr/src/run_new.sh
```

and re-run the recipe, the new image is pulled, and run, BUT with the old command.
When I run `docker ps` I see '/usr/src/run.sh' as the command, but expect '/usr/src/run_new.sh'. Yet the image is the new one, containing other updates if things other than the CMD are updated.

Eibwen:
### Description

When a dockerfile has a command at the end, if that command in the dockerfile is modified chef will not respect that change.

Full description from my coworker who made the change:

> The container resource uses a common pattern of loading the current running values on the server, so that it can compare them to the values specified in our resource definition to see if anything has changed and know if it needs to restart the container with new values (https://github.com/chef-cookbooks/docker/blob/master/libraries/docker_container.rb#L122). The docker API call returns all properties of a running container, regardless of whether they were specified in the previous chef client run. The problem is that one of those properties is the "Command", which we actually don't specify currently in our resource definition, relying on the "CMD" property in the dockerfile to take care of that for us.
> 
> So, in our scenario, when the container resource runs, it grabs the Command property from the running container (which is the Command from the previous version of the dockerfile). It detects that something has changed in the resource (the "tag" property), so it knows it needs to stop, delete, and re-run the container. Since we don't specify the Command property in our resource definition, it passes the currently running value of Command into the docker API, along with all the other properties (https://github.com/chef-cookbooks/docker/blob/master/libraries/docker_container.rb#L237). So we end up with a container that looks like it's been upgraded, but is actually running the command from the previous version of the dockerfile.
### Issues Resolved
#673
### Check List
- [x] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD

…in resource

### DIFF
--- a/libraries/docker_container.rb
+++ b/libraries/docker_container.rb
@@ -234,7 +234,7 @@ module DockerCookbook
             'name'            => container_name,
             'Image'           => "#{repo}:#{tag}",
             'Labels'          => labels,
-            'Cmd'             => to_shellwords(command),
+            'Cmd'             => property_is_set?(:command) ? to_shellwords(command) : nil,
             'AttachStderr'    => attach_stderr,
             'AttachStdin'     => attach_stdin,
             'AttachStdout'    => attach_stdout,


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.getchef.com/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## GitHub Users Who Are Not Authorized To Contribute

The following GitHub users do not appear to have signed a CLA:

* @clearmeasurebrian

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

Eibwen:
Any help to figure out exactly how to write tests around this new function would be appreciated.  The `load_current_value` function did not seem to be called by any of the current tests, so we didn't have examples to follow.

chasebolt:
hmm have you tested this change? instinctively i think it wouldnt work because of load_current_value setting the property.

chasebolt:
i believe this is resolved now. extensively tested with latest chef/docker/compat_resource. closing for now. if you still see it happening please open an issue and provide a test case - if possible.

PR #744, opened 2016-08-24
Minor updates

tas50:


tas50:
New chefdk with the fix for the travis runs should be in current tomorrow.

PR #759, opened 2016-10-10
make systemd MountFlags configurable

dy-dx:
### Description

Fixes https://github.com/chef-cookbooks/docker/issues/756
### Issues Resolved

https://github.com/chef-cookbooks/docker/issues/756
### Check List
- [x] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD
- [ ] New functionality includes testing.
- [x] New functionality has been documented in the README if applicable
- [x] The CLA has been signed. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD

PR #763, opened 2016-10-21
Allow customizing systemd MountFlags option

FlorentFlament:
### Description

This change allows customizing the Systemd MountFlags option when Docker is launched by systemd.
### Issues Resolved

Prior to that change, the Systemd MountFlags option was hardcoded to 'private' in the template file.


tas50:
I think #759 makes more sense since it includes the readme update as well. I'm going to close this PR out. Thanks for putting it together though.

PR #768, opened 2016-11-09
Added ro_rootfs => ReadonlyRootfs special cases mapping

yruss972:
### Description

This commit adds ReadonlyRootfs to the list of keys which are skipped in the automated load of state and then maps ro_rootfs to the ReadonlyRootfs manually as was done with other similar cases.

### Issues Resolved

The ro_rootfs property of the docker_container resource doesn't match the property in the docker HostConfig (called ReadonlyRootfs).
As a result, chef fails to properly load the value into the state of the current resource and forces a redeploy even when no change has been made.

### Check List

to prevent redeploying containers with ro_rootfs = true

Signed-off-by: Yonah Russ <me@yonahruss.com>


chasebolt:
LGTM thanks!

PR #787, opened 2016-12-14
When stopping via execute, wait until Docker is fully stopped

RoboticCheese:
### Description

I've only been able to reproduce this when doing Docker-in-Docker via
kitchen-docker, but there are cases where the restart can happen faster
than Docker can stop. There is a split second after killing the process
where the PID file still exists, and Docker will fail to start if the
PID file exists.

### Check List

Signed-off-by: Jonathan Hartman <j@p4nt5.com>


RoboticCheese:
The build failure was because of some RuboCop offenses unrelated to this change.

PR #815, opened 2017-01-31
vendoring docker-api-1.33.2

someara:


methodmissing:
Hi, just wondering about the Mach specific shared objects making it's way into the gem vendor as it'll be useless on Linux and other platforms ...

* https://github.com/chef-cookbooks/docker/pull/815/files#diff-22bc1cdd05c897ff1484caba0aceb4d1

* https://github.com/chef-cookbooks/docker/pull/815/files#diff-ca3854f43faa571c10f9c133cc2a6256

someara:
whoops! nice catch. I didn't realize json would get pulled in as a dep.
reverting.... going back to the autoload method in 2.14.3

PR #824, opened 2017-02-15
Require 12.5 and remove the need for compat_resource

tas50:
Avoid a pile of pain with compat_resource


someara:
YAAAASS
(will fix up travis before release)

PR #853, opened 2017-04-30
Add network_aliases support

jsirex:
### Description

Docker host_name actually doesn't set containers' hostname globally to
the docker network. But `--network-alias` does.


someara:
Thanks! Released in v2.15.6

PR #857, opened 2017-05-11
Remove support for Precise and Vivid

tas50:
These are both EOL and have old kernsl. Let them die

Signed-off-by: Tim Smith <tsmith@chef.io>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 sudo: required
 dist: trusty
 
-# install the pre-release chef-dk.  Use chef-stable-trusty to install the stable release
+# install the pre-release chef-dk.
 addons:
   apt:
     sources:
-      - chef-stable-trusty
+      - chef-current-trusty
     packages:
       - chefdk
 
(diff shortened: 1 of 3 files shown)


someara:
Gotta keep Presicse

tas50:
It's EOL and not getting security updates anymore.

someara:
It gets security updates into 2019 through Extended Security Maintenance.
Docker explicitly publishes precise packages, which will work by installing a modern kernel from backports.

-s

chasebolt:
the new docker repos, which i have been switching everything over to, no longer supports precise and vivid. https://download.docker.com/linux/ubuntu/dists/

i think we should drop support as it will make testing more difficult and i dont even want to think about the feature support once i get this library up to date with all the new docker features.

chasebolt:
ok ill keep the helper logic for precise. imo, if someone is going through all the effort in still using precise with a newer docker cookbook, they really should be using the `package_name` and `package_version` properties.

someara:
I resemble that remark!

chasebolt:
vivid has been removed. keeping precise until someone *cough* @someara *cough* upgrades a few systems.

PR #867, opened 2017-06-06
Bug: Change kill_after property default value to nil

tkilloren:
### Description

I was getting read timeout errors when using the :restart action on multiple docker_container resources in the same run.

```
           Docker::Error::TimeoutError
           ---------------------------
           read timeout reached

           Cookbook Trace:
           ---------------
           /tmp/kitchen/cache/cookbooks/docker/libraries/docker_container.rb:396:in `block (2 levels) in \<class:DockerContainer\>'
```

Noticed that the default value of the 'kill_after' property on the docker_container resource was '-1'.    When looking at the docker-api gem code (https://github.com/swipely/docker-api/blob/v1.33.1/lib/docker/container.rb#L229-L235) saw that this would now force set 'read_timeout' and 'write_timeout' to 4 seconds.   Looks like 'nil' would be the proper default value to use for the documented default behavior of never killing the container (with recent versions of the docker-api gem).

### Issues Resolved

No existing open issue found.

### Check List

### DIFF
--- a/libraries/docker_container.rb
+++ b/libraries/docker_container.rb
@@ -100,9 +100,9 @@ module DockerCookbook
     property :container, Docker::Container, desired_state: false
 
     # Used by :stop action. If the container takes longer than this
-    # many seconds to stop, kill itinstead. -1 (the default) means
+    # many seconds to stop, kill it instead. A nil value (the default) means
     # never kill the container.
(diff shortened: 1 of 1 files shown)


chasebolt:
thanks!

Issue #839, opened 2017-04-07
2.15.2 breaks upstart on Docker 1.8


CVTJNII:
### Cookbook version
2.15.2

### Chef-client version
2.19.36

### Platform Details
Ubuntu 14.04.1 LTS

### Docker Version
1.8.2

### Scenario:
Unable to start Docker

### Steps to Reproduce:
"service docker start" hangs as Docker can't start.  This also blocks convergence.

### Expected Result:
Docker should start

### Actual Result:
Docker doesn't start

This appears to be a breaking change between 2.6.1 and 2.15.2.  In 2.15.2 the --raw-logs flag was added to the daemon flags in /etc/init/docker.conf:

```
root@default-ubuntu-1404:~# grep raw.logs /etc/ -ir
/etc/init/docker.conf:    exec "$DOCKER" daemon $DOCKER_OPTS --raw-logs
```

This is not compatible with the older version of Docker, which is still supported per the readme (or at least 1.8.3).  We are pinned back to the older Docker version due to blocking Docker bugs.  It would be nice if this breaking change could be resolved in the same major version.  (I won't complain about dropping the older versions, but please bump major, and I think it's late for that.)  Thanks!

iennae:
Did #870 resolve this for you?

chasebolt:
closing  for now. please reopen if this is still an issue with the latest version.

PR #870, opened 2017-06-14
Only use --raw-logs argument in versions which support it

drags:
### Description

The upstart config in this cookbook includes the daemon argument `--raw-logs`, which is only valid in docker-engine versions after [1.11.0](https://github.com/moby/moby/blob/89658be/CHANGELOG.md#logging-6), however the argument is currently hard coded, breaking the docker daemon on any upstart using system running docker daemons prior to version 1.11.0.

This change adds a `docker_raw_logs_arg` helper which enables or disables the raw-logs argument depending on the version used. This changeset updates the upstart template to make use of the `docker_raw_logs_arg` helper.

### Issues Resolved

https://github.com/chef-cookbooks/docker/issues/839

### Check List

- [x] [N/A] New functionality has been documented in the README if applicable

---

Regarding testing: I'm unfamiliar with kitchen and InSpec. Can I get some guidance on where to add tests here? Is testing for older versions of docker even required?

### DIFF
--- a/libraries/docker_service_manager_upstart.rb
+++ b/libraries/docker_service_manager_upstart.rb
@@ -23,6 +23,7 @@ module DockerCookbook
           docker_name: docker_name,
           dockerd_bin_link: dockerd_bin_link,
           docker_daemon_arg: docker_daemon_arg,
+          docker_raw_logs_arg: docker_raw_logs_arg,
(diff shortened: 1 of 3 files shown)


chasebolt:
thanks!

Issue #882, opened 2017-07-06
docker_service_manager_upstart not started correctly after initial creation


eyalzek:
### Cookbook version
2.15.13

### Chef-client version
12.19.36

### Platform Details
ubuntu 14.04 (aws)

### Scenario:
creating a `docker_service_manager_upstart` with some custom configuration will not pick up said configuration on first run after the initial creation. Only on subsequent runs (after a restart of the service).

### Steps to Reproduce:
create the following resource:

```
docker_service_manager_upstart 'default' do
  action :start
  host 'unix:///var/run/docker.sock'
  dns '8.8.8.8'
  dns_search 'some.search.domain.com
  storage_driver 'aufs'
end
```

### Expected Result:
upstart service would be created and it should be running with the correct configuration as set on the resource.

### Actual Result:
`dockerd` service runs with the default distribution values, e.g:
`27808 ? Ssl 0:00 /usr/bin/dockerd --raw-logs`

restarting the service will cause the correct values to be picked up and from then on it works as expected.
In the meantime my fix to the issue is to add a `notifies` to the resource described above:

```
docker_service_manager_upstart 'default' do
  action :start
  host 'unix:///var/run/docker.sock'
  dns '8.8.8.8'
  dns_search 'some.search.domain.com
  storage_driver 'aufs'
  notifies :restart, 'docker_service_manager_upstart[default]'
end
```

iennae:
@eyalzek Thanks for reporting!

chasebolt:
released in `v2.15.17`

PR #886, opened 2017-07-18
adding restart notifications to upstart

chasebolt:
adding restart notifications to upstart and cleaning house on the configs


chasebolt:
resolves #882

PR #887, opened 2017-07-19
Only restart docker.service once both docker.service and docker.socke…

dylancanfield:
…t overrides have been updated

### Description

There was a problem with the way that the systemd unit customizations for docker.service and docker.socket were applied. When changing attributes that affect the socket, namely the socket group, docker.service would be restarted before the overrides to the docker.socket unit were applied. This would cause docker.socket to fail to start and thus docker.service to fail to start.

This change just delays the restarting of docker.service and docker.socket to allow both overrides to be applied before attempting to restart these systemd units.

### Issues Resolved

[List any existing issues this PR resolves]

### Check List

### DIFF
--- a/libraries/docker_service_manager_systemd.rb
+++ b/libraries/docker_service_manager_systemd.rb
@@ -69,7 +69,7 @@ module DockerCookbook
         )
         cookbook 'docker'
         notifies :run, 'execute[systemctl daemon-reload]', :immediately
-        notifies :run, "execute[systemctl restart #{docker_name}]", :immediately
+        notifies :run, "execute[systemctl restart #{docker_name}]", :delayed
         action :create
       end
 
@@ -86,7 +86,7 @@ module DockerCookbook
         )
         cookbook 'docker'
         notifies :run, 'execute[systemctl daemon-reload]', :immediately
-        notifies :run, "execute[systemctl restart #{docker_name}]", :immediately
+        notifies :run, "execute[systemctl restart #{docker_name}]", :delayed
         action :create
       end
 


chasebolt:
can you please gist the logs of this happening? `docker.service` doesn't have anything unique in it except for `docker_name` and shouldnt be restarted on a socket property change.

AlvinFeng:
Would this fix the issue I have here https://github.com/chef-cookbooks/docker/issues/888?

chasebolt:
@AlvinFeng your issue is unrelated to this PR

dylancanfield:
@AlvinFeng, docker.service depends on docker.socket. However, docker.service is started before docker.socket overrides are applied via `templates/default/systemd/docker.socket-override.erb`. This template for the socket does not allow for much customization, but does allow you to change the group which owns the socket. In some cases, where the default group, `docker` does not exist, docker.socket fails to start without applying the override and thus docker.service fails to start and the converge fails before fixing the problem. This change simply delays the starting of both services and does not change the restart dependency between them.

Here is an example on RHEL7 which causes the failure:
```
package 'docker' do                                                             
  action :install                                                               
end

docker_service node['cub_docker']['service_name'] do                            
  group 'dockerroot'                                   
  action [:create, :start]                                                      
end
```

In this case, we are using the docker package maintained by RedHat. It uses the dockerroot group, rather than the docker group. The docker group simply does not exist.

Here is a gist of the chef output and a few command from inside the VM
https://gist.github.com/dylancanfield/6a4ea21295a4ca4cbbd93d368b1d8c2f

I realize that RHEL isn't officially supported by this cookbook, however, this bug will affect anyone who is running docker under a different group as the first converge will start the service and socket with different groups.

chasebolt:
a delayed restart on docker service will fail when creating any containers that rely on changes to the docker service

dylancanfield:
@chasebolt I had not considered that. Perhaps the correct way to do this is to explicitly call a restart to #{docker_name}.socket when the socket definition is updated and do that before calling a restart of the service.

e.g.

```
template "/etc/systemd/system/#{docker_name}.socket" do
  ...
  notifies :run, 'execute[systemctl daemon-reload]', :immediately         
  notifies :run, "execute[systemctl restart #{docker_name}.socket]", :immediately
  action :create
end

template "/etc/systemd/system/#{docker_name}.service" do   
  ...
  notifies :run, 'execute[systemctl daemon-reload]', :immediately         
  notifies :run, "execute[systemctl restart #{docker_name}]", :immediately
  action :create
end
```

Thoughts?

I have already tested this approach in our environment and it passes all of our integration tests (for what that's worth). I can submit a new PR unless you have objections with that approach.

Also, thank you for your time.

chasebolt:
@dylancanfield can you give release `v2.15.18` a spin and see if it resolves all of your issues?

dylancanfield:
@chasebolt Yes release `v2.15.18` resolves all of our issues. Thank you for publishing those changes so quickly.

Issue #876, opened 2017-06-30
docker_network should support ipv6, internal


rmoriz:
Looks like the current resource is not on a par with the current docker cli:

https://github.com/chef-cookbooks/docker/blob/master/libraries/docker_network.rb#L9-L20

vs:

```
docker -v
Docker version 17.05.0-ce, build 89658be

docker network create --help

Usage:	docker network create [OPTIONS] NETWORK

Create a network

Options:
      --attachable             Enable manual container attachment
      --aux-address map        Auxiliary IPv4 or IPv6 addresses used by Network driver (default map[])
  -d, --driver string          Driver to manage the Network (default "bridge")
      --gateway stringSlice    IPv4 or IPv6 Gateway for the master subnet
      --help                   Print usage
      --ingress                Create swarm routing-mesh network
      --internal               Restrict external access to the network
      --ip-range stringSlice   Allocate container ip from a sub-range
      --ipam-driver string     IP Address Management Driver (default "default")
      --ipam-opt map           Set IPAM driver specific options (default map[])
      --ipv6                   Enable IPv6 networking
      --label list             Set metadata on a network
  -o, --opt map                Set driver specific options (default map[])
      --subnet stringSlice     Subnet in CIDR format that represents a network segment
```

Especially `ipv6` and `internal`…

chasebolt:
resolved via #889

PR #889, opened 2017-07-20
docker_network: support ipv6

rmoriz:
### Description

allow setting the [EnableIPv6](https://docs.docker.com/engine/api/v1.30/#operation/NetworkCreate) option to create networks with IPv6 support.

### DIFF
--- a/libraries/helpers_network.rb
+++ b/libraries/helpers_network.rb
@@ -92,7 +92,6 @@ module DockerCookbook
             data[s]['Gateway'] = g
             match = true
           end
-          raise "no matching subnet for gateway #{s}" unless match
         end
 
         auxaddrs.each do |aa|
--- a/test/cookbooks/docker_test/recipes/network.rb
+++ b/test/cookbooks/docker_test/recipes/network.rb
@@ -225,3 +225,28 @@ docker_network 'network_h1 disconnector' do
   network_name 'network_h1'
   action :disconnect
 end
+
+##############
+# network_ipv6
+##############
+# IPv6 enabled network
+docker_network 'network_ipv6' do
+  enable_ipv6 true
+  subnet 'fd00:dead:beef::/48'
+  action :create
+end
+
+##############
+# network_ipv4
+##############
+docker_network 'network_ipv4' do
+  action :create
+end
+
+##################
+# network_internal
+##################
+docker_network 'network_internal' do
+  internal true
+  action :create
+end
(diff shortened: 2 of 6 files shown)


rmoriz:
issue #876

test failure caused by unrelated change (Fedora: No candidate version available for docker-engine) :(

rmoriz:
(maybe) blocked by issue in #890 - I think there should be no requirement to specify a gateway address but currently it does PR does not work without setting one 😒

test/cookbooks/docker_test/recipes/network.rb
```diff
@@ -225,3 +225,19 @@
   network_name 'network_h1'
   action :disconnect
 end
+
+###########
+# network_i
+###########
+
+# IPv6 enabled network
+docker_network 'network_i1' do
```
chasebolt: rename to `network_ipv6`

test/cookbooks/docker_test/recipes/network.rb
```diff
@@ -225,3 +225,19 @@
...
+  enable_ipv6 true
+  subnet 'fd00:dead:beef::/48'
+  action :create
+end
+
+# IPv6 disabled network (default)
+docker_network 'network_i2' do
```
chasebolt: rename to `network_ipv4`

test/cookbooks/docker_test/recipes/network.rb
rmoriz: My intent was to keep the naming consistant with the Docker API https://docs.docker.com/engine/api/v1.30/#operation/NetworkCreate - the docker network CLI option is "--ipv6".

rmoriz:
Okay, let me also add the "internal" network flag

libraries/helpers_network.rb
```diff
@@ -92,7 +92,6 @@ def consolidate_ipam(subnets, ranges, gateways, auxaddrs)
             data[s]['Gateway'] = g
             match = true
           end
-          raise "no matching subnet for gateway #{s}" unless match
```
chasebolt: is there a gateway being set? still confused on how this is being triggered if `gateways` is empty.

rmoriz: the inspect looks like:

```json

 "IPAM": {
            "Driver": "default",
            "Options": null,
            "Config": [
                {
                    "Subnet": "172.24.0.0/16",
                    "Gateway": "172.24.0.1"
                },
                {
                    "Subnet": "fd00:dead:beef::/48"
                }
            ]
        },
```
Docker automatically adds the IPv4 part (Gateway set) but the IPv6 subnet never matches a Gateway.

chasebolt: good enough for me. thanks for the clarification.

rmoriz:
sadly this still has mayor issues with idempotency. And I must say this type of library-cookbook is super hard to understand :(

PR #892, opened 2017-07-24
Support correctly centos as well

kamaradclimber:
### Description

Correctly default to dockerroot group on all rhel family. Previous patch was working only on redhat

### Check List

### DIFF
--- a/libraries/helpers_service.rb
+++ b/libraries/helpers_service.rb
@@ -46,7 +46,7 @@ module DockerCookbook
       end
 
       def default_group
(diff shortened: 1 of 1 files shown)


chasebolt:
using the yum docker repo on centos installs the group as `docker`. i dont have access to redhat and so far it seems redhat users are installing docker from redhat - so it is defaulted to `dockerroot` for compatibility.

this cookbook is tested against (and recommends) the yum docker repo. if you want to stick with centos version of docker you'll need to set the property `group 'dockerroot'` on your service.

PR #895, opened 2017-07-27
fix to_bytes parsing

rmoriz:
regex in https://github.com/chef-cookbooks/docker/blob/ebfbb1598bfc45dff8c6ec025893d22a9633cef3/libraries/helpers_container.rb#L21 is broken and only splits lower-case characters.


chasebolt:
thanks!

PR #896, opened 2017-07-27
stable docker version

chasebolt:
the original docker repos we have been testing against are now docker-experimental builds. this bumps to the latest stable docker version in the repos.

fedora 26 is currently not supported by docker. versioning to fedora-25.

before this PR is merged in chef-apt-docker cookbook needs to be updated. chef-yum-docker is already complete.

https://github.com/chef-cookbooks/chef-yum-docker/pull/7


someara:
Let's keep the helper logic but "remove support" by dropping it from the test matrix.

PR #905, opened 2017-08-24
Stop and start upstart service instead of restarting

akarasik:
### Description

When changing a /etc/init/<service>.conf file, restart won't get the new config.
Only running stop, and then start will get the latest config from disk.

http://upstart.ubuntu.com/cookbook/#restart


someara:
thanks! released in 2.15.25

PR #907, opened 2017-08-25
If the /etc/sysconfig file changes, restart Docker

hynd:
Signed-off-by: Kieren Hynd <kieren.hynd@ticketmaster.co.uk>

### Description
Similar to some of the other service providers, if the RHEL sysvinit provider changes the /etc/sysconfig/docker file, restart the service.  Otherwise if we're using docker_service to set any docker_daemon_opts, they don't get picked up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 This file is used to list changes made in each version of the docker cookbook.
 
+## 2.15.27 (2017-08-25)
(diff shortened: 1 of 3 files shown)


chasebolt:
thanks! released in 090f6a6638bd98b8c9a62cf3d9597837679007ff. just a heads up for your next PR dont commit to the changelog and metadata. most releases include multiple PRs. thanks again :)

PR #914, opened 2017-09-12
Fail on deprecations in Travis

tas50:
Clear up all our deprecations


tas50:
This is going to fail, but will allow me to fix all the stuff

PR #928, opened 2017-10-03
add env_file back into docker_container fixes #872

### LINKED ISSUE #872: [QUESTION] docker env_file
Does this cookbook support the docker `env_file` parameter?

enkaskal:


chasebolt:
docker run has a —env-file, we would want to pass that through the config json instead of reading in a file and passing individual variables.

enkaskal:
@chasebolt, I thought this cookbook was refactored away from shelling out to `docker run` with the 1.x rewrite. Further, I looked through https://github.com/swipely/docker-api and https://docs.docker.com/engine/api/v1.32/#operation/ContainerCreate and I don't see support for EnvFile. Can you please point me in the right direction?

enkaskal:
accidentally closed earlier when commenting and one of the tests failed because the node couldn't receive a key from ubuntu's keyserver. therefore, closing and reopening again as advised by: https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit/18252580

chasebolt:
released `env_file` support in v2.16.1

PR #946, opened 2017-12-09
Update Amazon Linux package to 17.06.2ce-1.94.amzn1

### LINKED ISSUE #945: Docker 17.06.2ce-1.94.amzn1 available on Amazon Linux 2017.09
### Cookbook version
v2.17.0

### Chef-client version
12.21.26

### Platform Details
Amazon Linux 2017.09

### Scenario:
`docker-17.06.2ce-1.94.amzn1` is available on Amazon Linux from the amzn-updates repository but the cookbook is still defaulting to `docker-17.06.2ce-1.93.amzn1`.

Any chance we can get a default docker version bump in helpers_installation_package.rb?

davidcaste:
### Description

Package `docker-17.06.2ce-1.94.amzn1` is available on Amazon Linux from the amzn-updates repository but the cookbook is still defaulting to `docker-17.06.2ce-1.93.amzn1`.

This PR attempts to update the default package version installed on Amazon Linux.

### Issues Resolved

#945

### DIFF
--- a/libraries/helpers_installation_package.rb
+++ b/libraries/helpers_installation_package.rb
@@ -109,7 +109,7 @@ module DockerCookbook
 
         return "#{v}#{edition}-1.el6" if el6?
         return "#{v}#{edition}-1.el7.centos" if el7?
-        return "#{v}#{edition}-1.93.amzn1" if amazon?
+        return "#{v}#{edition}-1.94.amzn1" if amazon?
         return "#{v}#{edition}-1.fc#{node['platform_version'].to_i}" if fedora?
         return "#{v}#{edition}-0~#{debian_prefix}#{codename}" if node['platform'] == 'debian'
         return "#{v}#{edition}-0~#{ubuntu_prefix}#{codename}" if node['platform'] == 'ubuntu'


et304383:
Can we consider rewriting the package install logic of this cookbook to just NOT try to install the Docker version specified if a newer one is already installed?  Having this cookbook randomly fail every time a new Docker version is published to the yum repositories is unacceptable.

tas50:
@et304383 I'd certainly like to move that way

et304383:
@tas50 perhaps this is an improvement for the package resource in Chef itself.

If a newer version of a package is installed than what is specified, I'd like to not error out.  Perhaps an action of "downgrade" would be appropriate.  Or a flag on whether or not it should be an exception if a newer version is installed compared to what version was requested.

tas50:
@et304383 So there's some back story here. The goal in this cookbook was to allow you to pin to minor version of docker, but the current yum package resource don't allow that. You can say: Install 1.2*. You have to specify 1.2.3-el-whatever-whatever. It leads to problems like this and we have a VERY large PR that fundamentally reworks how yum package installs work in the works right now. Once that lands we *could* fix this entirely, but doesn't that would require the very latest and greatest version of chef. It might come down to us requiring exactly that, but obviously we have to ship that fix first.

https://github.com/chef/chef/pull/6540

TLDR: We're entirely behind fixing this issue, but we have to completely refactor yum package in chef-client itself to do so

chasebolt:
@et304383 how are

chasebolt:
i agree - this is definitely a pain point and hopefully will get resolved in the future. im open to anything we can do to make it easier in the mean time.

et304383:
Well the issue comes up when a yum update is run against a system to update packages and Docker is one of those packages.

Though admittedly I've had this issue on other cookbooks that do version pinning for packages such as the MongoDB and Elasticsearch cookbooks.

I think it's a bigger overall issue that isn't easily solved.

chasebolt:
right. telling yum to update the system but then telling chef to make sure package X has version Z - this is definitely conflicting.

PR #949, opened 2017-12-13
Install docker-api with gem metadata. Requires Chef 12.10+

tas50:
Simplifies the gem install logic


chasebolt:
one issue i ran into this method when testing was that the system requires build tools for some of the gem dependencies. our systems dont have build tools installed out of the gate.

tas50:
It does break things when you have gems which need to be compiled, but luckily we don't have any here:

Fetching: excon-0.59.0.gem (100%)
Successfully installed excon-0.59.0
Fetching: multi_json-1.12.2.gem (100%)
Successfully installed multi_json-1.12.2
Fetching: docker-api-1.34.0.gem (100%)
Successfully installed docker-api-1.34.0
3 gems installed

chasebolt:
Awesome. I think it was a json gem that needed build tools at the time - glad it is resolved now.

chasebolt:
Yup they just fixed this a month ago https://github.com/swipely/docker-api/commit/4db485067d6b54f570d37cda7087d1279cc1f68f

PR #958, opened 2018-01-14
Expand testing and further cleanup legacy support

tas50:
Automatically setup the apt/yum repositories in the installation_package resource. Add new property for disabling this if you want to manage your own repo.
Fully remove Debian 7 / RHEL 6 support
Add support for Ubuntu 17.10
Fix the tarball installation to use the new Docker CE files
Allow specifying the channel in the tarball installer
Default tarball installs to 17.12 and remove old release SHAs
Remove old logic that was only necessary for EOL releases
Remove ChefSpec matchers that aren't needed in ChefDK 2+
Fix the remove action in docker_installation_script to remove the right package name
Remove turning on why run, which is the default in Chef 13
Replace deprecated --graph option with --data-root option
Use --host instead of -H so it's more obvious what we're doing
Test all resources in Travis instead of just the simple smoke test
Convert inspec test to use the native Inspec resources for Docker where we can (most places)
Remove failing specs that don't work with the latest Docker releases
Fix package installation on Fedora where Chef is doing the right thing now in DNF land


chasebolt:
can we keep one smoke test in there? it is nice for doing a quick kitchen test before pushing a PR.

lots of good changes thanks!

tas50:
@chasebolt @someara This is ready for a merge / major release

chasebolt:
tons of good changes thanks!

PR #975, opened 2018-03-14
Added oom_kill_disable and oom_score_adj support to docker_container

alrf:
### Description

Added **oom_kill_disable** and **oom_score_adj** support to docker_container.

https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources
https://docs.docker.com/config/containers/resource_constraints/#memory


alrf:
Issue: https://github.com/chef-cookbooks/docker/issues/976

tas50:
Thanks @alrf. This will get shipped in the next release and I'll make sure we pull it into chef-client when we move these resources

PR #977, opened 2018-03-14
Remove the TLS default helpers

tas50:
These aren't helpful.

Signed-off-by: Tim Smith <tsmith@chef.io>


lamont-granquist:
:+1: useless layers of indirection are useless

PR #1013, opened 2018-07-27
docker_execute -> docker_exec in README.md

caleb:
The example shows the correct resource name, but the prose says the name is `docker_execute`


tas50:
Thanks

PR #1020, opened 2018-08-16
Implement docker_plugin resource

### LINKED ISSUE #942: Add support for a docker_plugin resource
### Cookbook version
latest

### Chef-client version
irrelevant

### Platform Details
irrelevant

### Scenario:

I want to be able to config-manage installation docker plugins - my eventual intent is to be able to use the netshare volume driver in order to mount nfs shares as docker volumes.

### Steps to Reproduce:
Missing functionality.

### Expected Result:
Sucessfully install a docker volume driver, and then install an NFS volume.

### Actual Result:
Unsupported feature.

horazont:
### Description

Implement a ``docker_plugin`` resource which is used to manage [Docker Engine managed plugins](https://docs.docker.com/engine/extend/). This includes installation, enabling/disabling, (re-)configuring and removal of plugins.

### Issues Resolved

* Fixes #942 

### Check List

### Notes

* Waiting for travis-ci results on the "All tests pass" front :)
* I’m not entirely sure how to test this better. There are some smoke-testy-tests in `plugins.rb` + corresponding spec tests in the test cookbook, but I’m not 100% happy with it yet. I’m open to suggestions on how to improve this. I only found the two plugins listed (many other plugins are available, but not as Docker Engine managed ones), and both depend on external resources (an ssh server or a Ceph cluster), so creating volumes using those plugins doesn’t seem viable.
* As mentioned in #942, docker-api does not support any Plugin-specific stuff, so we’re using the raw connection object to communicate with Docker. This is even less overhead than I had anticipated, so I think it might be ok to leave it at this.

Let me know if anything is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ test/cookbooks/docker_test/
 - [docker_registry](#docker_registry): registry operations
 - [docker_network](#docker_network): network operations
 - [docker_volume](#docker_volume): volume operations
+- [docker_plugin](#docker_plugin): plugin operations
 
 ## Getting Started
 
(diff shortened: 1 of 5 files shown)


tas50:
@horazont Thanks for putting this all together. I'm out of the office for the rest of the week, but I'll dig into this next week. It's something we really want to get into this cookbook so we can bake it in and then move it right into chef-client.

PR #1022, opened 2018-08-17
Fix passing of driver and opts to Docker::Volume.create

horazont:
### Description

The `driver` and `opts` properties of the `docker_volume` resource were not used at all, despite their existence being documented in the README. This PR fixes this by passing the properties to the `Docker` call and documenting what they do and when they’re used.

### Issues Resolved

None

### Check List

### Notes

* I haven’t written tests for this. I have no idea how to do this. I could write something like:

   ```ruby
   docker_volume 'foo' do
     driver 'local'
     action :create
   end
   ```

   but that would not test anything useful, since ``local`` is the default on the Docker side of things, so not passing it and passing ``local`` has the same effect. A "good" way to test this would be with a separate volume driver, however, all drivers I know of require provisioning of an external resource (e.g. sshfs or ceph).

   However, I was able to create a ``wetopi/rbd:1.0.1`` based volume with this fix in our own infrastructure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1218,11 +1218,11 @@ The `docker_volume` resource is responsible for managing Docker named volumes.
 
 ### Properties
 
-- `driver`
+- `driver` - Name of the volume driver to use. Only used for `:create`.
 - `host`
-- `opts`
+- `opts` - Options to pass to the volume driver. Only used for `:create`.
 - `volume`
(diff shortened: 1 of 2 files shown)


tas50:
Thanks for noticing this and putting together the PR

PR #1026, opened 2018-08-23
Adding auto_restart. Default: true. service will restart on update un…

guyyosan:
…less this is set to false.

### Description

[Describe what this change achieves]

### Issues Resolved

[List any existing issues this PR resolves]

### Check List

### DIFF
--- a/libraries/docker_service_base.rb
+++ b/libraries/docker_service_base.rb
@@ -20,7 +20,7 @@ module DockerCookbook
 
     # daemon management
     property :instance, String, name_property: true, desired_state: false
-    property :auto_restart, [TrueClass, FalseClass], default: false
+    property :auto_restart, [TrueClass, FalseClass], default: true
     property :api_cors_header, String
     property :bridge, String
     property :bip, [IPV4_ADDR, IPV4_CIDR, IPV6_ADDR, IPV6_CIDR, nil]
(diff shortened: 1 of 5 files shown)


chasebolt:
we are required to restart docker on update to it will load our version of the systemd unit file. iirc restarting the docker service does not restart the containers.

guyyosan:
What if you update the dns for a kubernetes cluster?
For it to update the containers must be restarted...

libraries/docker_service_base.rb
```diff
@@ -20,7 +20,7 @@ class DockerServiceBase < DockerBase
 
     # daemon management
     property :instance, String, name_property: true, desired_state: false
-    property :auto_restart, [TrueClass, FalseClass], default: false
+    property :auto_restart, [TrueClass, FalseClass], default: true
```
majormoses: Not sure how I feel about making this change, IMO it is a breaking change and I dont think its a safe default. In most cases you probably want some additional orchestration on top to do this or it could mean that all your services could restart at the same time. I am supportive of people being able to opt in for this behavior but I disagree on this being the default behavior. Just my `$0.02`

guyyosan: @majormoses The auto_restart flag was configured as a property and had no logic in the code.
further more, I didn't change the default behaviour, which is a restart for the service on change of the template.

Only if you change the auto_restart to flase in the docker_service definition you'd get a service that never restarts on changes.

guyyosan: @chasebolt 
I know we are required to restart the service that's why I kept the default behaviour to do so.

And only in certain specific occasions the user can decide to change that, to have a docker service that doesn't restarts on changes.

tas50: This is actually a no-op since this property wasn't used before. The previous behavior was basically a true so this should be good to go.

scalp42:
@guyyosan why did you close this PR?

PR #1033, opened 2018-10-04
Import docker repository gpg key from http

gionn:
### Description

When installing docker via package on debian platforms, fetch via httsp the latest gpg public key server from download.docker.com

### Issues Resolved

Help avoiding keyserver unavailability transient errors, that is one of the most common transient network errors during first converge.

### Check List

### DIFF
--- a/libraries/docker_installation_package.rb
+++ b/libraries/docker_installation_package.rb
@@ -31,7 +31,7 @@ module DockerCookbook
             uri "https://download.docker.com/linux/#{node['platform']}"
             arch 'amd64'
             keyserver 'keyserver.ubuntu.com'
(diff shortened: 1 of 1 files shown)


tas50:
A cleaner alternative is probably to just put a retry on the apt_repository resource.

gionn:
@tas50 that was the first solution I think about, but then I realized that since we still rely on download.docker.com availability to fetch .deb packages we could prefer to pull the public key from there rather than a third-party keyserver that could be unreachable for more than a few retries.

tas50:
You can also just use URL in the key property and fetch it directly from docker. We don't need to grab this from a 3rd party key server.

tas50:
Thanks @gionn

Issue #998, opened 2018-06-08
Docker Container resource not idempotent with network_aliases defined


isuftin:
### Cookbook version
4.2.0

### Chef-client version
14.2.0

### Platform Details
CentOS 7 VM via Test Kitchen

### Scenario:
Set network alias on container via docker_container resource and not have it change during each Chef run

### Steps to Reproduce:
My test kitchen:
```
---
driver:
  name: vagrant
  provision: true
  customize:
    cpus: 1
    memory: 1024

provisioner:
  name: chef_zero

platforms:
  - name: centos-7

verifier:
  name: inspec

suites:
  - name: default
    run_list:
      - recipe[owi_docker::test_alias]

```

my test_alias recipe in my owi_docker cookbook:
```
docker_service 'default' do
  install_method 'package'
  version '18.03.1'
  ip_forward true
  ipv4_forward true
  ipv6_forward true
  package_name 'docker-ce'
  storage_driver 'overlay2'
  storage_opts [ "overlay2.override_kernel_check=true" ]
  mount_flags "private"
end

docker_network 'default_bridge'

docker_image 'nginx' do
  repo 'nginx'
  tag '1.13.8'
  action :pull
end

docker_container 'nginx' do
  repo 'nginx'
  tag '1.13.8'
  port [ '80:80', '443:443' ]
  network_mode 'default_bridge'
  network_aliases [ 'nginx' ]
  action :run
end

```

After the initial test kitchen run, I expect that subsequent runs will be idempotent.

Run 1 (after initial rest kitchen run that sets up all service, image pulls and running container)
```
 kitchen converge
-----> Starting Kitchen (v1.21.2)
-----> Converging <default-centos-7>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 7.0.3...
       Removing non-cookbook files before transfer
       Preparing validation.pem
       Preparing client.rb
-----> Chef Omnibus installation detected (install only if missing)
       Transferring files to <default-centos-7>
       Starting Chef Client, version 14.2.0
       resolving cookbooks for run list: ["owi_docker::test_alias"]
       Synchronizing Cookbooks:
         - owi_docker (0.0.13)
         - ohai (5.2.2)
         - docker (4.2.0)
       Installing Cookbook Gems:
       Compiling Cookbooks...
       Converging 4 resources
       Recipe: owi_docker::test_alias
         * docker_service[default] action create
           * docker_installation_package[default] action create
[...]

PR #1042, opened 2018-12-08
fix issues with network_mode

smcavallo:
### Description

Fix broken issues with existing network_mode implementation which will allow it to be functional.  
Should not break any existing functionality as this is not even called with the existing schema/api version of the docker api gem.
This only gets called when API is 1.21 or greater.  Unless you are overriding the api schema version or using a pre-release version of the docker-api gem this should have no impact.
(There is a blurb about this in the issue.)

It will be good to fix this now as they are preparing version 2.X of the docker api gem.
See - https://github.com/swipely/docker-api/tree/release-2.x
Once this 2.X gem is released and used, this cookbook may break without this change.

### Issues Resolved

https://github.com/chef-cookbooks/docker/issues/998

### Check List

- [ X] All tests pass. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD>
- [ X] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>

### DIFF
--- a/libraries/docker_container.rb
+++ b/libraries/docker_container.rb
@@ -423,11 +423,11 @@ module DockerCookbook
       ro_rootfs container.info['HostConfig']['ReadonlyRootfs']
       if container.info['NetworkSettings'] &&
          container.info['NetworkSettings']['Networks'] &&
-         container.info['NetworkSettings']['Networks'][new_resource.network_mode] &&
-         container.info['NetworkSettings']['Networks'][new_resource.network_mode]['IPAMConfig'] &&
-         container.info['NetworkSettings']['Networks'][new_resource.network_mode]['IPAMConfig']['IPv4Address']
+         container.info['NetworkSettings']['Networks'][network_mode] &&
+         container.info['NetworkSettings']['Networks'][network_mode]['IPAMConfig'] &&
+         container.info['NetworkSettings']['Networks'][network_mode]['IPAMConfig']['IPv4Address']
 
-        ip_address container.info['NetworkSettings']['Networks'][new_resource.network_mode]['IPAMConfig']['IPv4Address']
+        ip_address container.info['NetworkSettings']['Networks'][network_mode]['IPAMConfig']['IPv4Address']
       end
     end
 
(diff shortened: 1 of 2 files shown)


libraries/docker_container.rb
```diff
@@ -423,11 +423,11 @@ def to_snake_case(name)
       ro_rootfs container.info['HostConfig']['ReadonlyRootfs']
       if container.info['NetworkSettings'] &&
          container.info['NetworkSettings']['Networks'] &&
-         container.info['NetworkSettings']['Networks'][new_resource.network_mode] &&
```
tas50: Isn't this going to result in a unfound method error?

smcavallo: @tas50 - it shouldn't cause that error.
It actually used to be this way and was broken as part of this commit -
https://github.com/chef-cookbooks/docker/commit/b4bf74025af6ddd77e129e479dd19556e552d203

So - in effect - this would revert back to the older behavior, which oddly does work.  But - now that you mention it and I looked at it a bit deeper there is a better way of handling this.  
I'm going to revise this PR to what I think the original intention is when it was originally added in
https://github.com/chef-cookbooks/docker/commit/f4f477e61cb6510e4ceac2f03a5bd0d37c3c3d68

There is a better way to handle this - and the way of handling this should have tests.
I'll post a much longer explanation of the problem and a solution with tests shortly.

smcavallo:
@tas50 - here's the revised PR
This code is run during `load_current_value`
It looks like the intention is to only get the ip_address if it's set in NetworkSettings => Networks
I don't know why it's not required to capture the ip_address if that's not found.  I did not change that functionality but it could easily be added.

You can see the issue if you run the below commands against any container -
* Manually run `docker inspect` - you will see NetworkSettings => Networks in the output 

* Current docker API schema 1.16 - container.info['NetworkSettings']['Networks'] will be nil.
This is also why this code is almost never used today
Note that the IPAddress can be found in container.info['NetworkSettings']["IPAddress"] if we really want it.
```
container_name = 'rabbitmq'
require 'docker'
connection = Docker::Connection.new('unix:///var/run/docker.sock', {})
container = Docker::Container.get(container_name, {}, connection) 
irb(main):009:0> container.info['NetworkSettings']['Networks']
=> nil
```
* Docker API schema 1.21 - container.info['NetworkSettings']['Networks'] will be used!
Per https://docs.docker.com/engine/api/version-history/#v121-api-changes
`This field deprecates the.... NetworkSettings.IPAddress....which are still returned for backward-compatibility, but will be removed in a future version.`

So those of us who are overriding the docker API version WILL actually run this code.  And we'll want to get the IPAddress from the correct field.
```
Docker.send(:remove_const, 'API_VERSION')
Docker.const_set('API_VERSION', '1.21')
container_name = 'rabbitmq'
require 'docker'
connection = Docker::Connection.new('unix:///var/run/docker.sock', {})
container = Docker::Container.get(container_name, {}, connection) 
container.info['NetworkSettings']['Networks']
```
Hopefully this summarizes the what/why/how for anyone reading this.

Also hoping this long winded explanation makes sense.

tas50:
👍

PR #1058, opened 2019-02-01
fix: remove systemd override files

docwhat:
By providing the override files in `/etc/systemd/system/` it prevents the user from overriding any setting for `docker.service` or `docker.socket`.

The cookbook should just manage the files in `/lib/systemd/system`only.

### DIFF
--- a/libraries/docker_service_manager_systemd.rb
+++ b/libraries/docker_service_manager_systemd.rb
@@ -40,21 +40,6 @@ module DockerCookbook
         not_if { docker_name == 'default' && ::File.exist?('/lib/systemd/system/docker.service') }
       end
 
-      # this overrides the main systemd socket
-      template "/etc/systemd/system/#{docker_name}.socket" do
-        source 'systemd/docker.socket-override.erb'
-        cookbook 'docker'
-        owner 'root'
-        group 'root'
-        mode '0644'
-        variables(
-          config: new_resource,
-          docker_name: docker_name,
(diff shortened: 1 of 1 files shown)


chasebolt:
you can customize the systemd file to your hearts desire with this property https://github.com/chef-cookbooks/docker/blob/master/libraries/docker_service_base.rb#L70

docwhat:
@chasebolt said:
> you can customize the systemd file to your hearts desire with this property https://github.com/chef-cookbooks/docker/blob/master/libraries/docker_service_base.rb#L70

How do I get that to work for the `docker.socket` unit? It only seems to change `/etc/systemd/system/docker.service`

docwhat:
To add to the confusion, I can't use `edit_resource!` or `find_resource` to modify "template[/etc/systemd/system/docker.socket]".

My initial instinct was to add `/etc/systemd/system/docker.socket.d/override.conf` but that doesn't work if the file `/etc/systemd/system/docker.socket` exists.

My next instinct was to create `/etc/systemd/system/docker.socket` myself, but then I fight with this cookbook.

Then I tried to `edit_resource!(:template, '/etc/systemd/system/docker.socket')` but Chef can't find that resource.

Finally I tried `systemd_opt` but that doesn't work for `docker.socket`.

chasebolt:
fixed in #1059, removed the single hardcoded `SocketUser` by relying on default socket file

PR #1075, opened 2019-08-16
Allow to specify the IP when connecting to a docker network

sebastianriese:
### Description

This PR adds two properties to the `docker_network` resource: `ip` and `ip6`, which allow you to specify the IPv4 resp. IPv6 address the connected container will have in the network. This is useful in various settings (e.g. if a service in the container that is connected to multiple networks shall only bind to one of them, and therefore needs a predictable IP). If the IP does not match the current IP of the container, the container will be disconnected from the network and reconnected with the specified IP.

For extensibility and consistency with the chef-client code, the `ipaddr` library was replaced by `ipaddress`.

Note: If the docker-api gem changes the supported Docker API version, the code will have to be changed to use the new API parameters to set the IP addresses.

### Issues Resolved

–

### Check List


sebastianriese:
I ran the tests for the network resource locally with Chef 14 (and with the platform ubuntu 16.04), the CI is broken due to the accept license prompt that the latest Chef installers (Chef 15) make.

ramereth:
@sebastianriese can you please fix the issues with the delivery check?

sebastianriese:
Thanks for the reminder, all test except the release label validator pass now. Is there anything I should do about this check?

libraries/docker_network.rb
```diff
@@ -1,3 +1,5 @@
+require 'ipaddress'
```
ramereth, line 1: This should be moved into ``DockerCookbook::DockerNetwork``.

test/cookbooks/docker_test/recipes/network.rb
```diff
@@ -225,6 +225,59 @@
...
+  action :connect
+end
+
+docker_network 'network_i connector2b' do
+  container 'container2-network_i'
+  network_name 'network_i'
+  ip '172.30.0.12'
+  action :connect
+end
+
```
ramereth, line 280: Could you please add a test for IPv6 as well here?

ramereth (dismissed):
Can you please also update the CHANGELOG.md with an ``## Unreleased`` section?

libraries/docker_network.rb
sebastianriese: Could you point me to the style guide for sous-chef cookbooks that recommends this practise? [StackOverflow recommends to put require at the top](https://stackoverflow.com/a/605329), since the namespaces are imported to the toplevel anyway and it's the most common convention.

test/cookbooks/docker_test/recipes/network.rb
sebastianriese: Addressed in the newest verision.

sebastianriese:
> Can you please also update the CHANGELOG.md with an `## Unreleased` section?

Addressed in the newest version.

sebastianriese:
By the way, is it intentional the the CI just runs the smoke tests and the installation tests; shouldn't the full kitchen test suite be run at least for one platform? (Or am I just blind and overlooking things).

libraries/docker_network.rb
ramereth: We really don't have a style guide other than what `cookstyle` provides, however I was mostly looking at the style/standard this cookbook uses and trying to stick to that.

ramereth:
> By the way, is it intentional the the CI just runs the smoke tests and the installation tests; shouldn't the full kitchen test suite be run at least for one platform? (Or am I just blind and overlooking things).

I agree but unfortunately the integration tests do not work properly using kitchen-dokken and need to be run with kitchen-vagrant. Even then, it needs to be converged many times due to a variety of issues related to the amount of resources being tested at once.

I typically run those locally if any changes are impacted by those suites to ensure it works correctly.

libraries/docker_network.rb
sebastianriese: Okay, consistency with the surrounding code convinces me. Done.

ramereth:
Not sure if this is intentional or not bu I noticed when I converged the network suite a second time, the tests fail. Can you please make sure this isn't a bug in your code or test recipe additions? I'd expect it to work on every run.

damacus:
Thanks for the PR here @sebastianriese. Can you re-add that documentation to the documentation folder now it's been moved out of the main README please?

damacus:
Merging as is. The documentation can be added back later on.

damacus (approved):

kitchen-porter:
Released as: [9.4.0](https://github.com/sous-chefs/docker/releases/tag/9.4.0)

PR #1078, opened 2019-09-12
Add epoch to version string on RHEL and CentOS

### LINKED ISSUE #1065: Upgrading docker version
### Cookbook version
4.9.2

### Chef-client version
14.3

### Platform Details
Ubuntu 16.04

### Scenario:
Attempting to upgrade previously installed version of docker does not work.

### Steps to Reproduce:
Converge with `docker_installation_package` resource with `version` set to `17.12.1` and then converge again with `version` set to `18.09.2`.

### Expected Result:
Expected the 1st converge to install docker 17.12.1
Manually update version property
Expected the 2nd converge to upgrade docker to 18.09.2

### Actual Result:
The upgrade of docker does not happen.  The `package` resource called by `docker_installation_package` is marked up-to-date and skipped.

I believe this is happening because `docker_installation_package`'s use of the `package` resource is only setting the `action` to `:install` and not `:upgrade`.  Is there another way to go about upgrading the docker engine with this cookbook?

sean-nixon:
### Description

The docker_installation_package resource omits the "3" epoch number on versions >= 18.09 which causes yum's version comparison to get confused and think that a version is older than the installed version even if the primary version number if greater. This means you can't install one version and then upgrade it later. Instead you get a warning like the following:

```
yum_package[docker-ce] action install[2019-09-12T00:00:01+00:00] WARN:
yum_package[docker-ce] docker-ce has installed version 3:18.09.7-3.el7.x86_64,
which is newer than available version 19.03.2-3.el7. Skipping...)
```

Best I can tell, the epoch number is currently static at 3 in the docker-ce RPM spec (see https://github.com/docker/docker-ce-packaging/blob/master/rpm/SPECS/docker-ce.spec) so I've kept the same approach as with Debian/Ubuntu distros.

### Issues Resolved

#1065 (Partly anyway. There may be additional work on Ubuntu. Not much context for that issue)

### Check List

### DIFF
--- a/libraries/docker_installation_package.rb
+++ b/libraries/docker_installation_package.rb
@@ -175,7 +175,7 @@ module DockerCookbook
         return "5:#{v}~#{test_version}-0~debian-#{codename}" if debian?
         return "5:#{v}~#{test_version}-0~ubuntu-#{codename}" if ubuntu?
       elsif v.to_f >= 18.09 && el7?
-        return "#{v}-#{test_version}.el7"
+        return "3:#{v}-#{test_version}.el7"
       elsif v.to_f >= 18.09 && fedora?
         return v.to_s
       else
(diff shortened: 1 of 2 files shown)


bkonick:
I'm interested in getting this fixed. I accidentally submitted a PR (#1085) for the same problem as I didn't see this one.

sean-nixon:
Rebased from master and fixed unit tests. This is still an issue as far as I can tell so would love to see this merged.

tas50:
@sean-nixon I bumped the required version of Chef Infra Client to 15 in the latest release. This released improved how yum_package handles version strings and let's us pass versions like 19.03.8*. With that change the need for the brittle version logic went away for CentOS and Fedora so I'm going to close this one out now.

PR #1080, opened 2019-11-18
Default to version 19.03.5 and update/fix systemd unit

smcavallo:
Signed-off-by: S.Cavallo <smcavallo@hotmail.com>

### Description

- update to the latest version of docker for security reasons.  Should not by default install the pre-CVE-2019-5736 docker version.
- upgrade the systemd unit to more closely align with the docker provided default found at https://github.com/docker/docker-ce-packaging/blob/master/systemd/docker.service
This fixes the issue below where the containerd dependency is broken in newer versions of docker.

### Issues Resolved

https://github.com/chef-cookbooks/docker/issues/1062

### Check List

- [X ] All tests pass. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD>
- [X ] New functionality includes testing.
- [X ] New functionality has been documented in the README if applicable
- [X ] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>

PR #1082, opened 2019-11-18
Update format of docker tarball filenames > 18.06.3

smcavallo:
Signed-off-by: S.Cavallo <smcavallo@hotmail.com>

### Description

See - https://download.docker.com/linux/static/stable/x86_64/
Docker changed the format of the tarball filenames in versions > 18.06.3

This fixes the partial updates to fix the systemd unit as implemented in 
https://github.com/chef-cookbooks/docker/pull/1080
but was missing required updates in docker.service-override.erb 

Also - a recurring issue with the kitchen tests is that they fail the majority of the time and always for the container memory tests. The issue looks to be caused by attempting to launch containers with too little memory.  This PR increases the memory which has greatly improved the success rate.
I haven't seen any failures since the increase.  It should be possible to increase further as the build environment has quite a bit of memory which could be allocated ->  https://docs.travis-ci.com/user/reference/overview/

See - https://download.docker.com/linux/fedora/28/x86_64/stable/Packages/
docker changed the package names for fedora in 18.09 and above

### Check List

- [X ] All tests pass. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD>
- [X ] New functionality includes testing.
- [X ] New functionality has been documented in the README if applicable
- [X ] All commits have been signed for the Developer Certificate of Origin. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD>


.travis.yml
```diff
@@ -31,7 +31,7 @@ env:
   - INSTANCE=volume-ubuntu-1604
   - INSTANCE=resources-debian-9
   - INSTANCE=resources-centos-7
-  - INSTANCE=resources-fedora-31
```
tas50: it's building now. I forgot about that one. There is fedora-latest though which avoids the need for bumping things

smcavallo: great idea - just made this change

smcavallo:
@tas50 - i've implemented your suggestions - thanks again!  all tests pass now.   this PR is finally ready for review/consideration.

tas50:
Thanks for taking the time to get this all cleaned up and green again.

limitusus:
Just for note:
https://github.com/chef-cookbooks/docker/pull/1082/files#diff-896fd8158946f25225dc33e8f6d13754R4 adds a dependency to containerd.

My AmazonLinux2 environment failed to restart dockerd.service after updating cookbook, and after installing containerd RPM succeeded to start.

duncaan:
@limitusus - just provisioned a al2 instance, and tried to install docker from the default rpm repos and it installed containerd by default.  Could this be an older instance?

limitusus:
@duncaan 
I throughly understood that I was using one version older docker package.
The very latest one released in September has containerd dependency.
So on new systems there should be no issue.

Here is the two latest changelogs of docker RPM:

```
* Mon Sep 23 2019 Jacob Vallejo <jakeev@amazon.com> - 18.09.9ce-1
- Update packaging for 18.09.9ce
- Split container primitives into package dependencies (containerd, runc)

* Thu Jun 06 2019 Jamie Anderson <jamieand@amazon.com> - 18.06.1ce-10
- CVE-2018-15664
```

salsa-dev:
shit happens. this PR has broken many of the docker machines by introducing containerd

PR #1102, opened 2020-04-26
Use new_resource to read attribute

dud225:
### Description
This fixes the following error:
```
    NameError
    ---------
    undefined local variable or method `restart_policy' for #<#<Class:0x00000000054923c0>:0x0000000008167a80>

```


tas50:
Thanks @dud225

PR #1136, opened 2020-10-08
Sous Chefs Adoption

ramereth:
# Description

Describe what this change achieves

## Issues Resolved

List any existing issues this PR resolves

## Check List


metadata.rb
```diff
@@ -18,4 +18,4 @@
 supports 'ubuntu'
 
 chef_version '>= 15.0'
-gem 'docker-api', '~> 1.34.0'
```
tas50, line 21: This should be >= 1.34, < 3 probably so it works with older versions of chef correctly

Berksfile
```diff
@@ -2,6 +2,9 @@ source 'https://supermarket.chef.io'
 
 metadata
 
+# XXX: Temporary for testing
+cookbook 'etcd', github: 'sous-chefs/etcd', branch: 'sous-chefs-adoption'
```
ramereth, line 6: This will get removed before merging.

tas50 (dismissed):

libraries/docker_service_manager_systemd.rb
```diff
@@ -26,6 +26,16 @@ class DockerServiceManagerSystemd < DockerServiceBase
         not_if { docker_name == 'default' && ::File.exist?('/lib/systemd/system/docker.socket') }
       end
 
+      template '/etc/systemd/system/containerd.service' do
```
tas50, line 29: I don't think containerd goes all the way back to 17.12. You might want to figure out when it came out and then mark that as the oldest version of Docker supported.

ramereth: That thought had occurred to me. I'll look into that.

tas50:
Very nice cleanup. Thanks @ramereth

ramereth:
@tas50 still OK with a minor bump on this?

PR #1155, opened 2021-02-20
Add `buildargs` propety to the `docker_image` resource

### LINKED ISSUE #884: docker_image should support buildargs
Docker API v1.21 (release a long time ago) introduced support for buildargs (ARG):
https://docs.docker.com/engine/api/v1.21/#22-images

yet https://github.com/chef-cookbooks/docker/blob/master/libraries/docker_image.rb is missing the property.

m-zare:
# Description

Adds `buildargs` property to the `docker_image` resource.

The new property is supported in building from a Dockerfile, a directory, and a tarball.

## Issues Resolved

 #884 

## Check List

- [x] All tests pass. See TESTING.md for details.
- [x] New functionality has been documented in the README if applicable.

### DIFF
--- a/libraries/docker_image.rb
+++ b/libraries/docker_image.rb
@@ -16,7 +16,6 @@ module DockerCookbook
     property :rm, [true, false], default: true
     property :source, String
     property :tag, String, default: 'latest'
-    property :buildargs, [String, Array, Hash], coerce: proc { |v| coerce_buildargs(v) }
     property :buildargs, [String, Hash], coerce: proc { |v| v.is_a?(String) ? v : coerce_buildargs(v) }
 
     alias_method :image, :repo
--- a/libraries/docker_image.rb
+++ b/libraries/docker_image.rb
@@ -17,6 +17,7 @@ module DockerCookbook
     property :source, String
     property :tag, String, default: 'latest'
     property :buildargs, [String, Array, Hash], coerce: proc { |v| coerce_buildargs(v) }
+    property :buildargs, [String, Hash], coerce: proc { |v| v.is_a?(String) ? v : coerce_buildargs(v) }
 
     alias_method :image, :repo
     alias_method :image_name, :repo
@@ -28,14 +29,7 @@ module DockerCookbook
     ###################
 
     def coerce_buildargs(v)
-      case v
-      when Hash
-        format('{ %s }', v.map { |key, value| format('"%s": "%s"', key, value) }.join(','))
-      when Array
-        format('{ %s }', v.map { |label| format('"%s": "%s"', label.split(':').first, label.split(':').last) }.join(','))
-      else
-        v
-      end
+      "{ #{v.map { |key, value| "\"#{key}\": \"#{value}\"" }.join(', ')} }"
     end
 
     #########
--- a/libraries/docker_image.rb
+++ b/libraries/docker_image.rb
@@ -29,12 +29,13 @@ module DockerCookbook
 
     def coerce_buildargs(v)
       case v
-      when Hash
-        format('{ %s }', v.map { |key, value| format('"%s": "%s"', key, value) }.join(','))
-      when Array
-        format('{ %s }', v.map { |label| format('"%s": "%s"', label.split(':').first, label.split(':').last) }.join(','))
-      else
+      when Hash, nil
         v
+      else
+        Array(v).each_with_object({}) do |label, h|
+          parts = label.split(':')
+          h[parts[0]] = parts[1..-1].join(':')
+        end
       end
     end
 
--- a/libraries/docker_image.rb
+++ b/libraries/docker_image.rb
@@ -30,9 +30,9 @@ module DockerCookbook
     def coerce_buildargs(v)
       case v
       when Hash
-        '{ %s }' % v.map { |k, v| '"%s": "%s"' % [k, v] }.join(',')
+        format('{ %s }', v.map { |key, value| format('"%s": "%s"', key, value) }.join(','))
       when Array
-        '{ %s }' % v.map { |label| '"%s": "%s"' % [label.split(':').first, label.split(':').last] }.join(',')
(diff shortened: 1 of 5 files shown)


libraries/docker_image.rb
```diff
@@ -16,12 +16,37 @@ class DockerImage < DockerBase
...
+        str = ''
+        v.each do |key, value|
+          str += '"' + key.to_s + '": "' + value.to_s + '",'
+        end
+        '{ ' + str.delete_suffix(',') + ' }'
+      when Array
+        str = ''
+        Array(v).each do |label|
+          parts = label.split(':')
+          str += '"' + parts[0].to_s + '": "' + parts[1].to_s + '",'
```
bmhughes, line 42: This (and the Hash type) could be refactored to a map with a join at the end which wouldn't require stripping the extraneous comma at the end of the string as well. Using string interpolation rather than concat would be more readable too.
```suggestion
          v.map { |label| "\"#{label.split(':').first}\": \"#{label.split(':').last}\""}.join(',')
```

libraries/docker_image.rb
```diff
@@ -16,12 +16,37 @@ class DockerImage < DockerBase
...
+    ###################
+    # Property helpers
+    ###################
+
+    def coerce_buildargs(v)
+      case v
+      when Hash
+        str = ''
+        v.each do |key, value|
+          str += '"' + key.to_s + '": "' + value.to_s + '",'
```
bmhughes, line 35: As below but something like this instead:
```suggestion
          v.map { |k, v| "\"#{k}\": \"#{v}\""}.join(',')
```

bmhughes (changes requested):

kitchen-porter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a CHANGELOG entry.</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

bmhughes:
Definetly better, but I'm not personally a fan of using `format()` when you've got native string interpolation in Ruby (especially when you have to nest it for the array), I think escaping the double quotes in a interpolated string is the easier-to-read lesser evil. Thoughts @ramereth?

ramereth:
> Definetly better, but I'm not personally a fan of using `format()` when you've got native string interpolation in Ruby (especially when you have to nest it for the array), I think escaping the double quotes in a interpolated string is the easier-to-read lesser evil. Thoughts @ramereth?

I'm not really sure honestly
@tas50 @lamont-granquist do you have any recommendations or opinions?

damacus:
Firstly thanks for the contribution! 🎉 

Can we drop the use of a String here? Using a String, Hash & Array is making the code more complicated than we really need here for limited gain. 

How often are we going to use something other than a Hash?

m-zare:
@damacus, Thanks for the comment.
Actually `String` looks more natural from [Docker's point of view](https://docs.docker.com/engine/api/v1.41/#operation/ImageBuild). When it's `buildargs`, it looks for a string, even `Hash` and `Array` should be converted.
I passed `buildargs {"IMAGE_NAME"=>"busybox", "IMAGE_TAG"=>"latest"}`, Docker throws this error:
```
Docker::Error::ClientError: docker_image[image_5] (docker_test::image line 172) had an error: Docker::Error::ClientError: {"message":"error reading build args: invalid character '=' after object key"}
```

bmhughes:
@damacus means from a Chef resource point of view. Just have the resource property accept only a `Hash`, also the logic you've used is (to me at least) just really weird and not in the least bit elegant.

You can keep the Hash with `.map` as it'll return an Array which you can easily join into the required string for Docker with a one-liner.

bmhughes:
My suggestion would be:

**Resource**
```ruby
property :buildargs, [String, Hash], coerce: proc { |v| v.is_a?(String) ? v : coerce_buildargs(v) }
```

Then the coerce method can be as simple as:
```ruby
def coerce_buildargs(hash)
  hash.map { |arg, value| "#{arg}=#{value}" }.join(',')
end
```

ramereth:
I agree with @bmhughes suggestion

m-zare:
@bmhughes it makes sense.

> My suggestion would be:
> 
> **Resource**
> 
> ```ruby
> property :buildargs, [String, Hash], coerce: proc { |v| v.is_a?(String) ? v : coerce_buildargs(v) }
> ```
> 
> Then the coerce method can be as simple as:
> 
> ```ruby
> def coerce_buildargs(hash)
>   hash.map { |arg, value| "#{arg}=#{value}" }.join(',')
> end
> ```
> 
> (double check the docker build args format as that is off the top of my head)

bmhughes (dismissed):
LGTM

libraries/docker_image.rb
```diff
@@ -16,12 +16,22 @@ class DockerImage < DockerBase
     property :rm, [true, false], default: true
     property :source, String
     property :tag, String, default: 'latest'
+    property :buildargs, [String, Array, Hash], coerce: proc { |v| coerce_buildargs(v) }
+    property :buildargs, [String, Hash], coerce: proc { |v| v.is_a?(String) ? v : coerce_buildargs(v) }
```
ramereth, line 20: Are these supposed to be duplicated?

ramereth (changes requested):

libraries/docker_image.rb
m-zare: No, it's a mistake.

ramereth (approved):

kitchen-porter:
Released as: [7.7.0](https://github.com/sous-chefs/docker/releases/tag/7.7.0)

PR #1163, opened 2021-07-01
Fix `installed_docker_version` method on ppc64le which appends `v` to the version

ramereth:
At least on ppc64le, `docker --version` returns:

```
Docker version v20.10.7, build f0df350
```

Instead of the following on other platforms:

```
Docker version 20.10.7, build f0df350
```

This causes a problem when parsing with `Gem::Version` so this strips any `v` from the string.

Signed-off-by: Lance Albertson <lance@osuosl.org>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


detjensrobert (approved):

kitchen-porter:
Released as: [7.7.2](https://github.com/sous-chefs/docker/releases/tag/7.7.2)

PR #1185, opened 2021-09-15
Move the docker_exec resource to a library

damacus:
# Description

Move the docker_exec resource to a library

## Issues Resolved

NA

## Check List


bmhughes (approved):

Xorima (commented):
Should `host` be in it's own partial as we are using it everywhere or is that too much?

damacus:
> Should `host` be in it's own partial as we are using it everywhere or is that too much?

There's lots of stuff I want to refactor after the move to custom resources. Lots of the helper methods need to be moved into their own helper libraries rather than being shipped with the resources. 

...But I'm breaking these changes down into the smallest working chunk possible.

Xorima (approved):

kitchen-porter:
Released as: [9.2.0](https://github.com/sous-chefs/docker/releases/tag/9.2.0)

PR #1189, opened 2021-09-15
Remove the Docker version workaround

damacus:
# Description

Remove the Docker version workaround

## Issues Resolved

List any existing issues this PR resolves

## Check List


damacus:
I don't think this is needed anymore. According to the comment this was fixed *ages* ago. But we probably need some testing to see if this actually works!

chasebolt:
in chef are you able to specify just the version number now? years ago we couldn't and in order to specify a version you had to give the full string. users were opening tickets as they were unable to figure out how to specify a version. definitely maintaining this piece of code has been a pain.

user expected to do `version '20.10.7'`, but in actuality they had to do `version: '20.10.7~ce~3-0~debian'`

ramereth:
> in chef are you able to specify just the version number now? years ago we couldn't and in order to specify a version you had to give the full string. users were opening tickets as they were unable to figure out how to specify a version. definitely maintaining this piece of code has been a pain.
> 
> user expected to do `version '20.10.7'`, but in actuality they had to do `version: '20.10.7~ce~3-0~debian'`

I took a closer look at this and this is still the case. The `version` parameter is only used for the short version which figures out the full version string. You can also use the `package_version` parameter to do the full string yourself if you want to.

It looks like @tas50 removed the RHEL code earlier but that actually broke this since yum also requires the full version string such as `3:20.10.7-3.el7` which includes the epoch and the release version.

So I'd say instead of removing this, we should clean it up and fix it for RHEL again and make sure it works with Debian Bullseye.

damacus:
> in chef are you able to specify just the version number now? years ago we couldn't and in order to specify a version you had to give the full string. users were opening tickets as they were unable to figure out how to specify a version. definitely maintaining this piece of code has been a pain.
> 
> user expected to do `version '20.10.7'`, but in actuality they had to do `version: '20.10.7~ce~3-0~debian'`

Ah right yeah, I suspected this might be the case and that the comment was sort-of correct. But it totally wasn't. I'm happy to leave all this code in for now. 

Thanks everyone 😄

PR #1194, opened 2021-10-08
disable and stop docker.socket when stopping the service

evandam:
# Description

When stopping the docker service, the socket should also be stopped. Otherwise it can be inadvertently started by accessing the socket. 

## Issues Resolved

Currently there's unexpected behavior (to me, at least) when stopping/disabling the Docker service. I think this is likely systemd-specific behavior.

Steps:

1. Install Docker, stop/disable the service (testing on Ubuntu 18.04)

```ruby
docker_service 'default' do
  action [:create, :stop]
end
```

2. Reboot the node
3. Verify Docker isn't running

```bash
$ systemctl is-active docker.service
inactive
$ systemctl is-enabled docker.service
disabled
```

4. Run any `docker` command. Note the command will hang rather than saying it cannot connect to the Docker daemon.

```bash
$ docker ps
```

5. See that `docker.service` is now active

```bash
$ systemctl is-active docker.service
active
```

## Check List

- [x] A summary of changes made is included in the CHANGELOG under `## Unreleased`
- [x] New functionality includes testing. (I don't see a good place to test behavior stopping the service, but let me know).
- [x] New functionality has been documented in the README if applicable. (N/A I think).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
           - 'installation-script-test'
           - 'installation-package'
           - 'installation-tarball'
+          - 'install-and-stop'
           - 'smoke'
         exclude:
           - os: debian-9


Xorima:
Hi @evandam 

Thanks for this, could you add an entry in `CHANGELOG.md` under the `## Unreleased` heading with a bullet point of what you have done (Do not worry about versioning it, this is automatic) 

Also could you please put in some tests to ensure this is working and we do not accidentality break it in future, ideally with some comments or a link back to this PR for reference.

Thanks!

PR #1291, opened 2024-12-11
Docs: update docs for tmpfs

### LINKED ISSUE #1120: --tmpfs option missing for container resource
In some cases, using --tmpfs option for containers is the only way to avoid running the container in privileged mode. See https://docs.docker.com/storage/tmpfs/ for more info.

One example of an image that requires --tmpfs in order to run without privileged mode is that of Dell Boomi: https://hub.docker.com/r/boomi/atom

damacus:
Resolves #1120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


kitchen-porter:
Released as: [11.8.2](https://github.com/sous-chefs/docker/releases/tag/11.8.2)

